PR #5, opened 2019-03-27
Add commonly-used filesystem helper to utils

mjcarroll:
This is a file that is used in pluginlib, but could be convenient to have in other places. 

It includes the appropriate location for `std::filesystem` if it is supported, otherwise it has a small `path` object that implements the most basic functionality. 

https://github.com/ros/pluginlib/blob/1a4de29fa55173e9b897ca8ff57ebc88c047e0b3/pluginlib/include/pluginlib/impl/filesystem_helper.hpp

Connects to ros2/rclcpp#665

### DIFF
--- a/include/rcpputils/filesystem_helper.hpp
+++ b/include/rcpputils/filesystem_helper.hpp
@@ -33,7 +33,7 @@
 // This file is originally from:
 // https://github.com/ros/pluginlib/blob/1a4de29fa55173e9b897ca8ff57ebc88c047e0b3/pluginlib/include/pluginlib/impl/filesystem_helper.hpp
 
-/// Includes std::filesystem and aliases the namespace to `pluginlib::impl::fs`.
+/// Includes std::filesystem and aliases the namespace to `rcpputils::fs`.
 /**
  * If std::filesystem is not available the necessary functions are emulated.
  */
@@ -73,8 +73,11 @@ public:
   {}
 
   path(const std::string & p)  // NOLINT(runtime/explicit): this is a conversion constructor
-  : path_(p), path_as_vector_(split(p, std::string(1, preferred_separator)))
-  {}
+  : path_(p), path_as_vector_(split(p, preferred_separator))
+  {
+    std::replace(path_.begin(), path_.end(), '\\', preferred_separator);
+    std::replace(path_.begin(), path_.end(), '/', preferred_separator);
+  }
 
   std::string string() const
   {
--- a/include/rcpputils/filesystem_helper.hpp
+++ b/include/rcpputils/filesystem_helper.hpp
@@ -33,9 +33,11 @@
 // This file is originally from:
 // https://github.com/ros/pluginlib/blob/1a4de29fa55173e9b897ca8ff57ebc88c047e0b3/pluginlib/include/pluginlib/impl/filesystem_helper.hpp
 
-/// Includes std::filesystem and aliases the namespace to `rcpputils::fs`.
 /**
  * If std::filesystem is not available the necessary functions are emulated.
+ *
+ * Note: Once std::filesystem is supported on all ROS2 platforms, this class
+ * can be deprecated in favor of the built-in functionality.
  */
 
 #ifndef RCPPUTILS__FILESYSTEM_HELPER_HPP_
--- a/include/rcpputils/filesystem_helper.hpp
+++ b/include/rcpputils/filesystem_helper.hpp
@@ -1,4 +1,4 @@
-// Copyright (c) 2017, Open Source Robotics Foundation, Inc.
+// Copyright (c) 2019, Open Source Robotics Foundation, Inc.
 // All rights reserved.
 //
 // Software License Agreement (BSD License 2.0)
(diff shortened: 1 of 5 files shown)


include/rcpputils/filesystem_helper.hpp
```diff
@@ -0,0 +1,195 @@
...
+#include <vector>
+
+#ifdef _WIN32
+#define RCPPUTILS_IMPL_OS_DIRSEP '\\'
+#else
+#define RCPPUTILS_IMPL_OS_DIRSEP '/'
+#endif
+
+#ifdef _WIN32
+  #include <io.h>
```
thomas-moulard: Here and a few other places are not indented properly according to https://google.github.io/styleguide/cppguide.html#Preprocessor_Directives 
```
The hash mark that starts a preprocessor directive should always be at the beginning of the line.
```


It's also not consistent with the style of the rest of the file, i.e.

```
#ifdef FOO
# define BAR
#endif
```

test/test_filesystem_helper.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+#include "rcpputils/filesystem_helper.hpp"
+
+using path = rcpputils::fs::path;
+
+TEST(TestFilesystemHelper, join_path)
+{
+  auto p = path("foo") / path("bar");
+
+#ifdef _WIN32
```
thomas-moulard: To minimize the risk of having Windows-only compilation errors, I'd suggest doing the following:

```c++
#ifdef WIN32
static constexpr const bool is_win32 = true;
#else
static constexpr const bool is_win32 = false;
#endif

// [...]

TEST(...) {
  // => now both sides of what was an #ifdef are built on Linux so no code paths are left untested.
  if (is_win32) {
    // ...
  } else {
    // ...
  }
}
```

This is a very minor issue, but that'd be great to avoid ending up with ifdef's everywhere once we start writing more tests.

include/rcpputils/filesystem_helper.hpp
```diff
@@ -0,0 +1,195 @@
...
+{
+namespace fs = std::experimental::filesystem;
+}  // namespace rcpputils
+
+#  define RCPPUTILS__FILESYSYEM_HELPER__HAS_STD_FILESYSTEM
+# endif
+#endif
+
+// The standard library does not provide it, so emulate it.
+#ifndef RCPPUTILS__FILESYSYEM_HELPER__HAS_STD_FILESYSTEM
```
thomas-moulard: I understand the desire to use the standard library as much as we can. However, this is expanding the test matrix quite a bit. We have now multiple competing implementation of the same class and, in the end, we can't use more than what provide the emulation layer anyway.

Naive question: wouldn't it be better to just remove lines 44 to 80, use the "emulation layer" everywhere and directly switch to use the real thing once it's actually non-experimental and included in a C++ standard we target?

mjcarroll: Yes, ultimately the best case scenario would be to use `std::filesystem` everywhere, but macOS is still lagging in it's support of that.  I believe that even with C++17, `std::filesystem` isn't present.

I found that this header (or functionality like it) is used in a few places so that people could essentially "polyfill" filesystem functionality in their code.

Generally, I was hoping to avoid having to maintain our own variant of `std::filesystem`, but it appears that it won't be an option until macOS catches up.

mjcarroll:
> Thanks Michael!
> 
> A few things I want to raise there:
> 
> * this repo has been created by our team (AWS Robotics) which is why it requires 2 reviews, etc. We're trying to iterating internally on how we want to manage code reviews there. Hope it's ok!

Makes perfect sense.

> * I reviewed this code as it was "new". ROS 2 is a good reason to try to tackle tech debt so I hope this makes sense to you. Let us know if you think this is going in the right direction.

I think that it's probably fair to do so, this code has been hanging around pluginlib and friends for some time.

> 
> Please bare with us while we get used to review ROS 2 PRs!

include/rcpputils/filesystem_helper.hpp
mjcarroll: One option would be to directly wrap what is available in `rcutils` with some more convenient C++ functionality for the time being, so that we don't have to maintain two complete implementations.

test/test_filesystem_helper.cpp
mjcarroll: Good call!

mjcarroll:
CI to verify cross-platform:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6549/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2888/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5366/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6347/)

thomas-moulard:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6565/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2904/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5382/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6366/)

include/rcpputils/filesystem_helper.hpp
thomas-moulard: Thanks for the explanation - let's keep this aligned with the STL and hopefully we'll be able to remove this at some point.

include/rcpputils/split.hpp
```diff
@@ -0,0 +1,73 @@
+// Copyright (c) 2017, Open Source Robotics Foundation, Inc.
```
thomas-moulard: nit: 2019

mjcarroll:
Looks like some warnings on Windows, I'll try and get those fixed up before a merge.

thomas-moulard:
@mjcarroll Awesome - we're good to go once we fix those.

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6575/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2913/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5391/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6382/)

mjcarroll:
That `static constexpr` member was giving some issues in CI (But not locally?):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6578/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2916/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5394/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6386/)

thomas-moulard:
@mjcarroll looking at the logs it's because the `constexpr` symbol was never declared.
`constexpr` is like `inline` in the sense that it's only an "hint" for the compiler. Depending on the flags the compiler can remove the symbol altogether (and you won't have linker issues) or not.
Probably would need to have:

```
# foo.h
class X {
static constexpr const bool x = true;
};

# foo.cc
/* static */ constepxr const bool X::x = true;
```

I realized this example is not exactly what happened to you but there is something around those lines going on IMHO...

Anyway the last iteration is good to go :+1:

mjcarroll:
Yes, @sloretz and I were discussing this.  It looks like you can explicitly fix it in C++17 with `inline constexpr`.  I was trying to avoid adding the `cc` file and corresponding library, as we were header-only up to this point.

PR #10, opened 2019-04-08
Adds CODEOWNERS file

thomas-moulard:
Ensures that any two members of the @ros2/dev or @ros2/aws-robotics
teams can approve changes.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,1 @@
+* @ros2/aws-robotics-code-owners


nuclearsandwich:
Does this also trigger review request notifications for these teams which are "participating level" rather than bulk GitHub notifications? In which case you're sending higher priority notifications to a quite large number of individuals who, as far as I know, have no way to opt out of them for this repository.

thomas-moulard:
Ok it's wrong, I'm going to update this PR.

thomas-moulard:
@nuclearsandwich Trying out using a dedicated team for now. We'll see how it goes.

thomas-moulard:
@tfoote @emersonknapp PTAL

tfoote:
I believe the approval gating is related to having commit access. The codeowners is related to notifications/direct requests so I think this works as it it. If you want more reviews they can manually be requested.

PR #13, opened 2019-06-08
Add function 'find_and_replace'

jacobperron:
This will be used for improvements related to https://github.com/ros2/rmw_fastrtps/issues/277.

### DIFF
--- a/include/rcpputils/find_and_replace.hpp
+++ b/include/rcpputils/find_and_replace.hpp
@@ -34,6 +34,9 @@ find_and_replace(const std::string & input, const std::string & find, const std:
   std::string output = input;
   const std::size_t find_len = find.length();
   const std::size_t replace_len = replace.length();
+  if (find == replace) {
+    return output;
+  }
   if (0u == find_len) {
     return output;
   }
--- a/include/rcpputils/find_and_replace.hpp
+++ b/include/rcpputils/find_and_replace.hpp
@@ -1,0 +1,50 @@
+// Copyright 2019 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RCPPUTILS__FIND_AND_REPLACE_HPP_
+#define RCPPUTILS__FIND_AND_REPLACE_HPP_
+
+#include <string>
+
+namespace rcpputils
+{
+
+/// Find and replace all instances of a string with another string.
+/**
+ * \param[in] input The input string.
+ * \param[in] find The substring to replace.
+ * \param[in] replace The string to substitute for each occurance of `find`.
+ * \return A copy of the input string with all instances of the string `find` replaced with the
+ *   string `replace`.
+ */
+inline std::string
+find_and_replace(const std::string & input, const std::string & find, const std::string & replace)
+{
+  std::string output = input;
+  const std::size_t find_len = find.length();
+  const std::size_t replace_len = replace.length();
+  if (0u == find_len) {
+    return output;
+  }
+  std::size_t pos = output.find(find);
+  while (pos != std::string::npos) {
+    output.replace(pos, find_len, replace);
+    pos = output.find(find, pos + replace_len);
+  }
+  return output;
+}
+
+}  // namespace rcpputils
+
+#endif  // RCPPUTILS__FIND_AND_REPLACE_HPP_
--- a/include/rcpputils/find_and_replace.hpp
+++ b/include/rcpputils/find_and_replace.hpp
@@ -24,7 +24,7 @@ namespace rcpputils
 /**
  * \param[in] input The input string.
  * \param[in] find The substring to replace.
- * \param[in] replace The string to substitute for each occurance of `find`.
+ * \param[in] replace The string to substitute for each occurrence of `find`.
  * \return A copy of the input string with all instances of the string `find` replaced with the
  *   string `replace`.
  */
--- a/include/rcpputils/find_and_replace.hpp
+++ b/include/rcpputils/find_and_replace.hpp
@@ -57,11 +57,35 @@ find_and_replace(
   return output;
 }
 
-inline
-std::string
-find_and_replace(const std::string & input, const std::string & find, const std::string & replace)
+namespace detail
 {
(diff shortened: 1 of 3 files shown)


jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7329/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3496/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6005/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7152/)

test/test_find_and_replace.cpp
```diff
@@ -0,0 +1,70 @@
...
+  {
+    auto ret = rcpputils::find_and_replace("foo", "", "bar");
+    EXPECT_EQ("foo", ret);
+  }
+  // Empty replace
+  {
+    auto ret = rcpputils::find_and_replace("foo", "foo", "");
+    EXPECT_EQ("", ret);
+  }
+  // Single occurance
```
prajakta-gokhale: Minor nit:

s/occurance/occurrence

test/test_find_and_replace.cpp
```diff
@@ -0,0 +1,70 @@
...
+  // Multiple occurances
+  {
+    auto ret = rcpputils::find_and_replace("foobarfoobar", "foo", "baz");
+    EXPECT_EQ("bazbarbazbar", ret);
+  }
+  {
+    auto ret = rcpputils::find_and_replace("foofoobar", "foo", "baz");
+    EXPECT_EQ("bazbazbar", ret);
+  }
+  // find == replace
```
prajakta-gokhale: Suggestion: This case can be handled separately in the function itself to avoid an unnecessary iteration of the string.

include/rcpputils/find_and_replace.hpp
```diff
@@ -0,0 +1,50 @@
...
+ * \return A copy of the input string with all instances of the string `find` replaced with the
+ *   string `replace`.
+ */
+inline std::string
+find_and_replace(const std::string & input, const std::string & find, const std::string & replace)
+{
+  std::string output = input;
+  const std::size_t find_len = find.length();
+  const std::size_t replace_len = replace.length();
+  if (0u == find_len) {
```
prajakta-gokhale: We can also check if `0u == input.length()` here.

include/rcpputils/find_and_replace.hpp
```diff
@@ -0,0 +1,50 @@
...
+
+/// Find and replace all instances of a string with another string.
+/**
+ * \param[in] input The input string.
+ * \param[in] find The substring to replace.
+ * \param[in] replace The string to substitute for each occurance of `find`.
+ * \return A copy of the input string with all instances of the string `find` replaced with the
+ *   string `replace`.
+ */
+inline std::string
```
wjwwood: I'd recommend making this a template function, which works with `std::basic_string` instead. That way it's more flexible, but also that way you can avoid ODR violation without using `inline`.

test/test_find_and_replace.cpp
jacobperron: Fixed in 3fb4961

test/test_find_and_replace.cpp
jacobperron: Good point. Done in dec7808

include/rcpputils/find_and_replace.hpp
jacobperron: Added template function in eb1ec92. I still have an inline function for `std::string` so that we can handle raw strings. Let me know if there's a better way to handle this.

include/rcpputils/find_and_replace.hpp
jacobperron: Would this improve anything? We won't enter the while-loop anyways.

include/rcpputils/find_and_replace.hpp
wjwwood: Raw strings? You mean like `const char *`?

include/rcpputils/find_and_replace.hpp
```diff
@@ -0,0 +1,69 @@
...
+/// Find and replace all instances of a string with another string.
+/**
+ * \param[in] input The input string.
+ * \param[in] find The substring to replace.
+ * \param[in] replace The string to substitute for each occurrence of `find`.
+ * \return A copy of the input string with all instances of the string `find` replaced with the
+ *   string `replace`.
+ */
+template<
+  class CharT,
```
wjwwood: If you default the `CharT` to be `char`, then I think raw string will work? Not sure if this is the "right" thing to do, but I don't see how it would hurt.

include/rcpputils/find_and_replace.hpp
jacobperron: Yes. Basically, most usages in the tests won't compile.

include/rcpputils/find_and_replace.hpp
jacobperron: Doesn't seem to change anything. This [SO post](https://stackoverflow.com/a/20758753) suggests doing something like:

```
template<typename CharT, std::size_t Length>
void find_and_replace( const CharT (&string)[Length] )
{
    find_and_replace(std::string(string));
}
```

I'm not sure if that's more correct.

wjwwood: Hmm. I'll play around with it quickly. If I don't see anything we can merge it as-is, but the whole point was to get rid of the `inline`.

wjwwood: I pushed a change that I think will work, your solution didn't work because you needed a different `Length` template argument for each literal.

include/rcpputils/find_and_replace.hpp
```diff
@@ -0,0 +1,93 @@
...
+}  // namespace detail
+
+template<typename InputT, typename FindT, typename ReplaceT>
+auto
+find_and_replace(
+  InputT && input,
+  FindT && find,
+  ReplaceT && replace)
+{
+  auto input_str = detail::normalize_to_basic_string(input);
```
wjwwood: There is one limitation here which is that the three argument normalize to the same char typed `basic_string`. Which I think is fine.

jacobperron: Yeah, I think that's okay.

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7346/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3511/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6024/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7168/)

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7348/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3513/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6026/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7170/)

PR #14, opened 2019-06-11
use iterators for split

Karsten1987:
I am extending the current split implementation for a iterator based design. This allows to use the same functionality for split to be used in different containers (e.g. `std::list`, `std::set`) and gives the opportunity for custom iterators. Especially the latter seems quite useful in order to correctly split and parse a message definition and extract it as a tuple.

CI: (build: up-to rcpputils, test: packages-select rcpputils)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7342/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3507/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6020/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7163/)

Signed-off-by: Karsten Knese <karsten@openrobotics.org>

### DIFF
--- a/include/rcpputils/split.hpp
+++ b/include/rcpputils/split.hpp
@@ -36,12 +36,42 @@
 #ifndef RCPPUTILS__SPLIT_HPP_
 #define RCPPUTILS__SPLIT_HPP_
 
+#include <iterator>
 #include <sstream>
 #include <string>
 #include <vector>
 
 namespace rcpputils
 {
+
+/// Split a specified input into tokens using a delimiter and a type erased insert iterator.
+/**
+ * The returned vector will contain the tokens split from the input
+ *
+ * \param[in] input the input string to be split
+ * \param[in] delim the dlelimiter used to split the input string
+ * \param[in] insert iterator pointing to a storage contianer
+ */
+template<
+  class InsertIterator,
+  typename std::enable_if<
+    std::is_same<
+      InsertIterator &,
+      decltype(std::declval<InsertIterator>().operator=(std::declval<std::string>()))>::value
+  >::type * = nullptr>
+inline void
+split(const std::string & input, char delim, InsertIterator & it, bool skip_empty = false)
+{
+  std::stringstream ss;
+  ss.str(input);
+  std::string item;
+  while (std::getline(ss, item, delim)) {
+    if (skip_empty && item == "") {
+      continue;
+    }
+    it = item;
+  }
+}
 
 /// Split a specified input into tokens using a delimiter.
 /**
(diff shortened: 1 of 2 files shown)


include/rcpputils/split.hpp
```diff
@@ -36,34 +36,57 @@
...
  * \param[in] input the input string to be split
  * \param[in] delim the dlelimiter used to split the input string
- * \return A vector of tokens.
+ * \param[in] insert iterator pointing to a storage contianer
  */
-inline std::vector<std::string>
-split(const std::string & input, char delim, bool skip_empty = false)
+template<
+  class InsertIterator,
+  typename std::enable_if<
```
thomas-moulard: Could static_assert be used here instead to improve readability?

include/rcpputils/split.hpp
```diff
@@ -36,34 +36,57 @@
...
+    std::is_same<
+      InsertIterator &,
+      decltype(std::declval<InsertIterator>().operator=(std::declval<std::string>()))>::value
+  >::type * = nullptr>
+inline void
```
thomas-moulard: inline is not needed anymore here

test/test_split.cpp
```diff
@@ -221,3 +228,73 @@ TEST(test_split, split_backslash) {
     EXPECT_EQ("world", ret[5]);
   }
 }
+
+TEST(test_split, iterator)
```
thomas-moulard: This could be three test cases.

test/test_split.cpp
```diff
@@ -221,3 +228,73 @@ TEST(test_split, split_backslash) {
...
+    return *this;
+  }
+
+  std::tuple<std::string, std::string, std::string> get() const
+  {
+    return std::tie(package_name, middle_element, message_name);
+  }
+};
+
+TEST(test_split, custom_iterator)
```
thomas-moulard: It'd be nice to also have a test case for the error case with EXPECT_THROW.

include/rcpputils/split.hpp
Karsten1987: The reason I opted for SFINAE is that `static_assert` doesn't participate in function overloading. That being, the current template approach has the benefit of having another function declaration set with `std::enable_if` which accepts a different form of iterator. One of them could be just a class member function, e.g. `add` which takes the string and operates on it.
Does this make sense?

Karsten1987:
I hopefully addressed all your comments. I am happy to re-iterate over `static_assert` vs SFINAE. I don't necessarily feel super strong about it.

include/rcpputils/split.hpp
thomas-moulard: I see, that makes sense. I just feel it's hard to read and I wished there was a better way to write this, but I don't see any right now so this LGTM ;)

PR #16, opened 2019-07-11
Add endian helper until C++20

mjcarroll:
std::endian won't exist until C++20, and this helps remove a dependency on boost::endian in the meantime.

### DIFF
--- a/include/rcpputils/endian.hpp
+++ b/include/rcpputils/endian.hpp
@@ -1,0 +1,50 @@
+// Copyright 2019 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * If std::endian is not available, the necessary functions are emulated.
+ *
+ * Note: Once std::endian is supported on all ROS2 platforms, this class
+ * can be deprecated in favor of the built-in functionality.
+ *
+ * Note: std::endian is targeted for C++20
+ */
+
+#ifndef RCPPUTILS__ENDIAN_HPP_
+#define RCPPUTILS__ENDIAN_HPP_
+
+#ifdef __has_include
+#  if __has_include(<endian.h>)
+#    include <endian.h>
+#  endif
+#endif
+
+namespace rcpputils
+{
+// Ref: https://en.cppreference.com/w/cpp/types/endian#Possible_implementation
+enum class endian
+{
+#ifdef _WIN32
+  little = 0,
+  big    = 1,
+  native = little
+#else
+  little = __ORDER_LITTLE_ENDIAN__,
+  big    = __ORDER_BIG_ENDIAN__,
+  native = __BYTE_ORDER__
+#endif
+};
+}  // namespace rcpputils
+
+#endif  // RCPPUTILS__ENDIAN_HPP_
--- a/include/rcpputils/endian.hpp
+++ b/include/rcpputils/endian.hpp
@@ -33,6 +33,8 @@
 namespace rcpputils
 {
 // Ref: https://en.cppreference.com/w/cpp/types/endian#Possible_implementation
+// From: https://en.cppreference.com/w/Cppreference:FAQ, this is licensed
+// Creative Commons Attribution-Sharealike 3.0 Unported License (CC-BY-SA)
 enum class endian
 {
 #ifdef _WIN32
--- a/include/rcpputils/endian.hpp
+++ b/include/rcpputils/endian.hpp
@@ -24,6 +24,21 @@
 #ifndef RCPPUTILS__ENDIAN_HPP_
 #define RCPPUTILS__ENDIAN_HPP_
 
+// TODO(anyone) replace this macro when the appropriate C++20 value lands.
+#if !defined(RCPPUTILS_NO_STD_ENDIAN) && (__cplusplus <= 201703L)
+#define RCPPUTILS_NO_STD_ENDIAN
+#endif
+
+#if !defined(RCPPUTILS_NO_STD_ENDIAN)
+#include <type_traits>
+namespace rcpputils
+{
+using std::endian;
+}  // namespace rcpputils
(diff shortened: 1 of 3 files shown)


mjcarroll:
Verifying that this works on target platforms:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7533/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3670/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6170/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7362/)

mjcarroll:
Once more with some diagnostic output:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7534/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3671/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6171/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7363/)

mjcarroll:
Amended to fix the lint error.

test/test_endian.cpp
```diff
@@ -0,0 +1,61 @@
...
+{
+  // A platform should be one or other other
+  EXPECT_TRUE((rcpputils::endian::little == rcpputils::endian::native) ||
+    (rcpputils::endian::big == rcpputils::endian::native));
+}
+
+TEST(test_endian, runtime_endianness)
+{
+  // Compare the runtime evaluation to the predefined constant value.
+  EXPECT_TRUE(isLittleEndian() &&
```
clalancette: Nit; it doesn't matter for the test to succeed, but it makes a little more sense to me to first test whether the platform is little-endian, then run the test.

test/test_endian.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+  if (rcpputils::endian::little == rcpputils::endian::native) {
+    std::cout << "Compiler reports: little endian" << std::endl;
+  } else if (rcpputils::endian::big == rcpputils::endian::native) {
+    std::cout << "Compiler reports: big endian" << std::endl;
+  } else {
+    std::cout << "Compiler reports: mixed/unknown endian" << std::endl;
+  }
+
+  if(isLittleEndian()) {
```
clalancette: Nit: space after `if`.

test/test_endian.cpp
```diff
@@ -0,0 +1,61 @@
...
+    std::cout << "Runtime reports: little endian" << std::endl;
+  } else {
+    std::cout << "Runtime reports: big endian" << std::endl;
+  }
+}
```
clalancette: Should we add an `EXPECT_TRUE` for runtime-endianness == compiler-endianness?  Otherwise, doing the work seems somewhat pointless for the test.

include/rcpputils/endian.hpp
```diff
@@ -0,0 +1,52 @@
...
+#ifndef RCPPUTILS__ENDIAN_HPP_
+#define RCPPUTILS__ENDIAN_HPP_
+
+#ifdef __has_include
+#  if __has_include(<endian.h>)
+#    include <endian.h>
+#  endif
+#endif
+
+namespace rcpputils
```
clalancette: We may want to wrap this whole thing in another `ifndef` guard, specifically for the C++20 include guard.  Then one day when we switch to C++20, it will just pick that up and use it.  There's an example of doing something like this over at https://github.com/tcbrindle/cpp17_headers/blob/master/include/stx/optional.hpp (also, we may want to add the ability to force-disable it, as that has proven helpful to me in the past).

test/test_endian.cpp
mjcarroll: Yeah, fixed in a force push

test/test_endian.cpp
mjcarroll: That should be what line 37 is doing.

include/rcpputils/endian.hpp
mjcarroll: Done in latest.

test/test_endian.cpp
mjcarroll: Clarified in latest.

mjcarroll:
If anyone is curious about general compiler compatibility, I experimented a bit with godbolt: https://godbolt.org/z/jsv_Sy

test/test_endian.cpp
```diff
@@ -0,0 +1,63 @@
...
+bool isLittleEndian()
+{
+  uint16_t number = 0x0001;
+  uint8_t * numPtr = reinterpret_cast<uint8_t *>(&number);
+  return numPtr[0] == 1;
+}
+
+TEST(test_endian, is_defined)
+{
+  // A platform should be one or other other
```
thomas-moulard: typo

test/test_endian.cpp
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  if (isLittleEndian()) {
+    std::cout << "Runtime reports: little endian" << std::endl;
+  } else {
+    std::cout << "Runtime reports: big endian" << std::endl;
+  }
+}
+
+#ifdef _WIN32
```
thomas-moulard: Maybe unnecessary? It's always better to limit the amount of arch-specific code and the previous test should work also in the win32 case unless I miss something?

mjcarroll:
Yes, I added a `static_assert(false)` in the non-c++20 block and tested with:
* clang 8.0.0 `-std=c++2a`
* gcc 8.3 `-std=c++2a`

Of note, msvc v19.21 with `/std:c++latest` still does not implement this feature.

test/test_endian.cpp
mjcarroll: True, eliminated that test.

test/test_endian.cpp
mjcarroll: Fixed, thanks.

mjcarroll:
Once more, since it's quick:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7537/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3674/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6174/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7366/)

include/rcpputils/endian.hpp
```diff
@@ -0,0 +1,67 @@
...
+#endif
+
+#if !defined(RCPPUTILS_NO_STD_ENDIAN)
+#include <type_traits>
+namespace rcpputils
+{
+using std::endian;
+}  // namespace rcpputils
+#define RCPPUTILS_HAVE_STD_ENDIAN 1
+#endif  // __cplusplus > 201703L
```
emersonknapp: Nit: this `#endif` comment is correct but doesn't match the opening `#if`

PR #17, opened 2019-07-26
Add function 'find_and_replace' 

jacobperron:
Backport #13


jacobperron:
@ros2/aws-robotics-code-owners 

In preparation for Dashing patch release 2, it would be great if the `dashing` branch could be tagged and released so this change is included.

PR #18, opened 2019-08-01
File extension addition for camera calibration parser

klintan:
Added the file extension helper needed for camera calibration parser library.

### DIFF
--- a/include/rcpputils/filesystem_helper.hpp
+++ b/include/rcpputils/filesystem_helper.hpp
@@ -146,7 +146,8 @@ public:
   std::string extension() const
   {
     auto fname = filename();
-    auto split_fname = split(fname.string(), std::string("."));
+    const char * delimiter = ".";
+    auto split_fname = split(fname.string(), *delimiter);
     return "." + split_fname.back();
   }
 
--- a/include/rcpputils/filesystem_helper.hpp
+++ b/include/rcpputils/filesystem_helper.hpp
@@ -122,6 +122,13 @@ public:
     return path_.empty() ? path() : *--this->cend();
   }
 
+  std::string extension() const
+  {
+    auto fname = filename();
+    auto split_fname = split(fname.string(), std::string("."));
+    return "." + split_fname.back();
+  }
+
   path operator/(const std::string & other)
   {
     return this->operator/(path(other));
--- a/include/rcpputils/filesystem_helper.hpp
+++ b/include/rcpputils/filesystem_helper.hpp
@@ -43,6 +43,9 @@
 #ifndef RCPPUTILS__FILESYSTEM_HELPER_HPP_
 #define RCPPUTILS__FILESYSTEM_HELPER_HPP_
 
+#include <sys/stat.h>
+#include <sys/types.h>
+
 #include <algorithm>
 #include <string>
 #include <vector>
@@ -86,22 +89,6 @@ public:
   std::string string() const
   {
     return path_;
-  }
-
-  inline bool create_directories(const path & p)
-  {
-    path p_built;
-
-    for (auto it = p.cbegin(); it != p.cend(); ++it) {
-      p_built /= *it;
-
-  #ifdef _WIN32
-      _mkdir(p_built.string().c_str());
-  #else
-      mkdir(p_built.string().c_str(), S_IRWXU | S_IRWXG | S_IROTH | S_IXOTH);
-  #endif
-    }
-    return true;
   }
 
   bool exists() const
@@ -184,6 +171,22 @@ private:
 inline bool exists(const path & path_to_check)
 {
   return path_to_check.exists();
+}
+
+inline bool create_directories(const path & p)
+{
+  path p_built;
+
+  for (auto it = p.cbegin(); it != p.cend(); ++it) {
+    p_built /= *it;
+
+#ifdef _WIN32
+    _mkdir(p_built.string().c_str());
+#else
+    mkdir(p_built.string().c_str(), S_IRWXU | S_IRWXG | S_IROTH | S_IXOTH);
+#endif
+  }
+  return true;
 }
 
 #undef RCPPUTILS_IMPL_OS_DIRSEP
--- a/include/rcpputils/filesystem_helper.hpp
+++ b/include/rcpputils/filesystem_helper.hpp
@@ -43,9 +43,6 @@
 #ifndef RCPPUTILS__FILESYSTEM_HELPER_HPP_
 #define RCPPUTILS__FILESYSTEM_HELPER_HPP_
 
-#include <sys/stat.h>
-#include <sys/types.h>
-
 #include <algorithm>
 #include <string>
 #include <vector>
@@ -58,8 +55,11 @@
 
 #ifdef _WIN32
 #  include <io.h>
+#  include <direct.h>
 #  define access _access_s
 #else
+#  include <sys/stat.h>
+#  include <sys/types.h>
 #  include <unistd.h>
 #endif
 
@@ -98,7 +98,7 @@ public:
 
   bool empty() const
   {
-    return path_ == "";
+    return path_.empty();
   }
 
   bool is_absolute() const
@@ -130,12 +130,11 @@ public:
     return path_.empty() ? path() : *--this->cend();
   }
 
-  std::string extension() const
+  path extension() const
   {
-    auto fname = filename();
     const char * delimiter = ".";
-    auto split_fname = split(fname.string(), *delimiter);
-    return "." + split_fname.back();
+    auto split_fname = split(this->string(), *delimiter);
+    return split_fname.size() == 1 ? path("") : path("." + split_fname.back());
   }
 
   path operator/(const std::string & other)
--- a/include/rcpputils/filesystem_helper.hpp
+++ b/include/rcpputils/filesystem_helper.hpp
@@ -54,8 +54,8 @@
 #endif
 
 #ifdef _WIN32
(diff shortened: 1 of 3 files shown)


thomas-moulard:
@klintan Could you please write tests and check this works on all platforms (OS X / Windows in particular)

klintan:
> @klintan Could you please write tests and check this works on all platforms (OS X / Windows in particular)

Added some tests.

Any tips on how to test it out on Windows ? are there any Windows 10 docker out there that can be used without a license? 

Is this running on Jenkins? I could create a circleci config for this one to run there, but not sure if that is what other core libs do? (similar to this https://github.com/ros-drivers/velodyne/blob/dashing-devel/.circleci/config.yml)

zmichaels11:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7885/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3937/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6432/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7752/)

test/test_filesystem_helper.cpp
```diff
@@ -103,3 +103,53 @@ TEST(TestFilesystemHelper, is_absolute)
     }
   }
 }
+
+TEST(TestFilesystemHelper, correct_extension)
```
thomas-moulard: it would be great to minimize the number of `is_win32` if-statements. This could be done as follow:

```c++
std::string build_example_path() {
  return is_win32 ? r"(C:\foo\bar\baz.yml)" : "/bar/foo/baz.yml";
}
```

...and then calling it in the tests directely.

test/test_filesystem_helper.cpp
```diff
@@ -103,3 +103,53 @@ TEST(TestFilesystemHelper, is_absolute)
...
+
+TEST(TestFilesystemHelper, is_empty)
+{
+  {
+    auto p = path("");
+    EXPECT_TRUE(p.empty());
+  }
+}
+
+TEST(TestFilesystemHelper, create_directories)
```
thomas-moulard: This should probably have a comment stating that this expected the current directory to be write-only (otherwise the test would fail). This is not great but I don't have an easy way to improve this.

test/test_filesystem_helper.cpp
```diff
@@ -103,3 +103,53 @@ TEST(TestFilesystemHelper, is_absolute)
...
+{
+  if (is_win32) {
+    {
+      auto p = path(".\\test_folder");
+      EXPECT_TRUE(rcpputils::fs::create_directories(p));
+    }
+  } else {
+    {
```
thomas-moulard: ditto

include/rcpputils/filesystem_helper.hpp
```diff
@@ -122,6 +130,14 @@ class path
     return path_.empty() ? path() : *--this->cend();
   }
 
+  std::string extension() const
```
thomas-moulard: This class is a drop-in replacement of the `std::filesystem::path` class. So we should match its interface: https://en.cppreference.com/w/cpp/filesystem/path/extension

If you really want a string, you could always make that a free funtion to avoid the problem.

thomas-moulard:
Additionally, your commit is missing a DCO.

thomas-moulard:
@klintan Let me know if you need some help to get this PR to the finish line - I have time this week so happy to help.

klintan:
> @klintan Let me know if you need some help to get this PR to the finish line - I have time this week so happy to help.

thank you so much, and thanks for the thorough review. I'll try to fix all the feedback tonight after work.

include/rcpputils/filesystem_helper.hpp
```diff
@@ -43,6 +43,9 @@
 #ifndef RCPPUTILS__FILESYSTEM_HELPER_HPP_
 #define RCPPUTILS__FILESYSTEM_HELPER_HPP_
 
+#include <sys/stat.h>
+#include <sys/types.h>
```
wjwwood: Are these POSIX specific? Maybe they belong below in the conditional block for includes.

https://stackoverflow.com/questions/25280440/why-include-direct-h-or-sys-stat-h-conditionally-based-on-win32-or-linux

include/rcpputils/filesystem_helper.hpp
wjwwood: +1, it should match the `std::filesystem` API. The `path` type has a `.string()` and `.c_str()`.

include/rcpputils/filesystem_helper.hpp
klintan: I fixed it, although, not sure `sys/types.h` is needed. Let me know if it looks ok :)

include/rcpputils/filesystem_helper.hpp
klintan: I updated it, it now returns empty string for no extension, and a path object for the extension. Let me know if it looks ok, I updated the tests as well.

test/test_filesystem_helper.cpp
klintan: I added some helper functions, not sure the placement is the best though, also the naming could perhaps be better. Let me know what you think of those and the placement etc.

thomas-moulard:
@ros2/aws-oncall @prajakta-gokhale @thomas-moulard - please run this CI job
Gist: https://gist.githubusercontent.com/thomas-moulard/01bc00a26f26ad6eab8b2ea39dbe4fcf/raw/5ce981dc958d5108f19d0023c422b4f0bfd22ae1/ros2.repos
BUILD args: --packages-up-to
TEST args: --packages-select
Job: ci_launcher

thomas-moulard:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7970/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3994/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6488/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7824/)

klintan:
Hm seems it failed on some cpplint stuff, I think i fixed it, but let me know if there are any other issues in there that I might have missed. 

Thank you so much @thomas-moulard for your patience and feedback! Really appreciate it

thomas-moulard:
Thanks! Retrying the CI

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7982/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4001/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6498/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7832/)

emersonknapp:
@klintan It looks like you've just got some whitespace problems remaining for the linters (e.g. https://ci.ros2.org/job/ci_linux/7982/testReport/junit/(root)/projectroot/cpplint/) - are you able to run the tests locally? I recommend `colcon test --packages-select rcpputils`, it shouldn't take very long.

Additionally, if you're not familiar, colcon mixins (https://github.com/colcon/colcon-mixin-repository) are very helpful, for example there's one called `linters-only`, e.g. `colcon test --packages-select rcpputils --mixin linters-only` - that command can be very helpful when you're just dealing with formatting and don't have the linters tied into your editor. Cheers!

klintan:
> colcon test --packages-select rcpputils

Thanks.  I didn't know about the mixins, looks really handy.

Yes I'm always running it locally, however for some reason the linters etc are running on all folders, even install/build/log folders, which then gives a bunch of errors for cmakefiles etc in those folders, so I sometimes misses the actual issues.
```
8: /Users/andreasklintberg/personal/rcpputils/build/rcpputils/CMakeFiles/feature_tests.c:0:  No copyright message found.  You should have a line: "Copyright [year] <Copyright Owner>"  [legal/copyright] [5]
```
For instance.

Do you know why that is ? Running `colcon test` in the base folder of the cloned rcpputils.

I'll fix the whitespaces tonight.

emersonknapp:
@klintan If i understand your message correctly, you've done

```
git clone rcpputils
cd rcpputils
colcon build
```

But what you should really be doing is creating a colcon workspace (along the lines of https://index.ros.org/doc/ros2/Installation/Crystal/Linux-Development-Setup/#id6) A minimal example

```
mkdir -p my_workspace/src
cd my_workspace
git clone rcpputils src/rcpputils
colcon build
colcon test
```

Again assuming my understanding was correct, your build created the `build`, `install`, `log` directories _inside_ the `rcpputils` repo directory, whereas the workspace configuration keeps them external, and therefore the linters won't see build artifacts.

Hope that helps!

emersonknapp:
Oops! Your last commit is missing the DCO. There are a lot of little things to pay attention to :sweat_smile: 

Once you fix that I will kick off the new CI and we should be good to go.

klintan:
> Oops! Your last commit is missing the DCO. There are a lot of little things to pay attention to 😅
> 
> Once you fix that I will kick off the new CI and we should be good to go.

Thanks for walking through the workspace issues.

Gah, yes sorry about that. Fixed!

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7990/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4010/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6508/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7841/)

PR #20, opened 2019-09-04
Implement join()

hidmic:
This pull requests introduces `join()`, the inverse of `split()`. Necessary for https://github.com/ros2/rclcpp/pull/842.

### DIFF
--- a/include/rcpputils/join.hpp
+++ b/include/rcpputils/join.hpp
@@ -1,0 +1,71 @@
+// Copyright (c) 2019, Open Source Robotics Foundation, Inc.
+// All rights reserved.
+//
+// Software License Agreement (BSD License 2.0)
+//
+// Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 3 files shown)


include/rcpputils/join.hpp
```diff
@@ -0,0 +1,71 @@
+// Copyright (c) 2019, Open Source Robotics Foundation, Inc.
+// All rights reserved.
+//
+// Software License Agreement (BSD License 2.0)
```
wjwwood: Was this copied from somewhere? Why the BSD License and not Apache 2?

hidmic: Oops, that's a bad copy-paste.

hidmic: Back to Apache 2 in https://github.com/ros2/rcpputils/pull/20/commits/1aceb4f6547fc3c6db95e7e63cf57b0006e92aa2.

PR #2, opened 2019-02-26
Add thread annotation macros

emersonknapp:
For clang thread safety analysis https://clang.llvm.org/docs/ThreadSafetyAnalysis.html

Related to https://github.com/ros2/ros2/issues/664

PR #22, opened 2019-09-05
Update README.md and package.xml

jacobperron:


package.xml
```diff
@@ -3,7 +3,7 @@
 <package format="2">
   <name>rcpputils</name>
   <version>0.1.0</version>
-  <description>Package containing utility code for C++. Currently only Clang thread safety annotation macros</description>
+  <description>Package containing utility code for C++.</description>
```
emersonknapp: Is this too little information? Do we want to say "... that isn't large enough to warrant a separate package" or something like that?

jacobperron: I could update it to match the C equivalent ([rcutils/package.xml](https://github.com/ros2/rcutils/blob/9af59e871e0bfaadb0cbb32014a7167efb9760c0/package.xml#L6))?
Feel free to suggest something better.

emersonknapp: I copied `rcutils` initially but was told it wasn't a very good description haha https://github.com/ros2/rcpputils/pull/2#discussion_r261014582 - i'm fine with it as is, just wanted to float the question

PR #28, opened 2019-12-18
Add utility function to get the size of a file

piraka9011:
## Changes
* Implemented a [`std::fs::file_size`](https://en.cppreference.com/w/cpp/filesystem/file_size) shim.
* Added a temporary directory fixture for GMock
* Add a unit test for `file_size()`.

Signed-off-by: Anas Abou Allaban <allabana@amazon.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,6 +54,7 @@ if(BUILD_TESTING)
   ament_add_gtest(test_split test/test_split.cpp)
 
   ament_add_gmock(test_filesystem_helper test/test_filesystem_helper.cpp)
+  target_link_libraries(test_filesystem_helper ${PROJECT_NAME})
 
   ament_add_gtest(test_find_and_replace test/test_find_and_replace.cpp)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,7 +24,8 @@ include_directories(include)
 ament_export_include_directories(include)
 
 add_library(${PROJECT_NAME}
-  src/find_library.cpp)
+  src/find_library.cpp
+  src/filesystem_helper.cpp)
 target_include_directories(${PROJECT_NAME}
   PUBLIC
   include
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,7 +52,7 @@ if(BUILD_TESTING)
 
   ament_add_gtest(test_split test/test_split.cpp)
 
-  ament_add_gtest(test_filesystem_helper test/test_filesystem_helper.cpp)
+  ament_add_gmock(test_filesystem_helper test/test_filesystem_helper.cpp)
 
   ament_add_gtest(test_find_and_replace test/test_find_and_replace.cpp)
 
--- a/include/rcpputils/filesystem_helper.hpp
+++ b/include/rcpputils/filesystem_helper.hpp
@@ -228,9 +228,8 @@ inline path remove_extension(const path & file_path, int n_times = 1)
  */
 inline std::uintmax_t file_size(const path & file_path)
 {
-  if (!file_path.exists())
-  {
-    std::error_code ec (ENOENT, std::system_category());
+  if (!file_path.exists()) {
+    std::error_code ec(ENOENT, std::system_category());
     throw std::system_error(ec, "Path does not exist");
   }
   struct stat stat_buffer {};
--- a/include/rcpputils/filesystem_helper.hpp
+++ b/include/rcpputils/filesystem_helper.hpp
@@ -45,6 +45,7 @@
 
 #include <algorithm>
 #include <string>
+#include <system_error>
 #include <vector>
 
 #ifdef _WIN32
@@ -215,6 +216,26 @@ inline path remove_extension(const path & file_path, int n_times = 1)
     new_path = path(new_path_str.substr(0, last_dot));
   }
   return new_path;
+}
+
+/**
+ * For a regular file p, returns the size determined as if by reading the st_size member of the
+ * structure obtained by stat.
+ * If p does not exist, reports an error.
+ *
+ * \param file_path Path to examine.
+ * \return The size of the file, in bytes.
+ */
+inline std::uintmax_t file_size(const path & file_path)
+{
+  if (!file_path.exists())
+  {
+    std::error_code ec (ENOENT, std::system_category());
+    throw std::system_error(ec, "Path does not exist");
+  }
+  struct stat stat_buffer {};
+  int rc = stat(file_path.string().c_str(), &stat_buffer);
+  return rc == 0 ? static_cast<size_t>(stat_buffer.st_size) : 0;
 }
 
 #undef RCPPUTILS_IMPL_OS_DIRSEP
--- a/include/rcpputils/filesystem_helper.hpp
+++ b/include/rcpputils/filesystem_helper.hpp
@@ -229,8 +229,9 @@ inline path remove_extension(const path & file_path, int n_times = 1)
 inline std::uintmax_t file_size(const path & file_path)
 {
   if (!file_path.exists()) {
-    std::error_code ec(ENOENT, std::system_category());
-    throw std::system_error(ec, "Path does not exist");
+    std::error_code ec{ENOENT, std::system_category()};
+    const std::string error_msg{"Path " + file_path.string() + " does not exist"};
+    throw std::system_error(ec, error_msg);
   }
   struct stat stat_buffer {};
   int rc = stat(file_path.string().c_str(), &stat_buffer);
--- a/include/rcpputils/filesystem_helper.hpp
+++ b/include/rcpputils/filesystem_helper.hpp
@@ -226,17 +226,7 @@ inline path remove_extension(const path & file_path, int n_times = 1)
  * \param file_path Path to examine.
  * \return The size of the file, in bytes.
  */
-inline std::uintmax_t file_size(const path & file_path)
-{
-  if (!file_path.exists()) {
-    std::error_code ec{ENOENT, std::system_category()};
-    const std::string error_msg{"Path " + file_path.string() + " does not exist"};
-    throw std::system_error(ec, error_msg);
-  }
-  struct stat stat_buffer {};
-  int rc = stat(file_path.string().c_str(), &stat_buffer);
-  return rc == 0 ? static_cast<size_t>(stat_buffer.st_size) : 0;
-}
+std::uintmax_t file_size(const path & file_path);
 
 #undef RCPPUTILS_IMPL_OS_DIRSEP
 
--- a/test/temporary_directory_fixture.hpp
+++ b/test/temporary_directory_fixture.hpp
@@ -17,17 +17,16 @@
 
 #include <gmock/gmock.h>
 
+#include <memory>
 #include <string>
 
 #ifdef _WIN32
-# include <direct.h>
-# include <Windows.h>
-#include <Shellapi.h>
-#include <shlobj.h>
+#include <direct.h>
+#include <Windows.h>
 #else
-# include <unistd.h>
-# include <sys/types.h>
-# include <dirent.h>
+#include <unistd.h>
+#include <sys/types.h>
+#include <dirent.h>
 #endif
 
 namespace
@@ -102,14 +101,14 @@ public:
     const auto null_terminated_directory = create_double_null_terminated_tchar(directory_path);
 
     SHFILEOPSTRUCT file_options{
-      .hwnd = nullptr;
-      .wFunc = FO_DELETE;  // Delete (recursively)
-      .pFrom = temp_dir;
-      .pTo = nullptr;
-      .fFlags = FOF_NOCONFIRMATION | FOF_SILENT;  // Do not prompt user
-      .fAnyOperationsAborted = FALSE;
-      .lpszProgressTitle = nullptr;
-      .hNameMappings = nullptr;
+      .hwnd = nullptr,
+      .wFunc = FO_DELETE,  // Delete (recursively)
+      .pFrom = temp_dir,
+      .pTo = nullptr,
+      .fFlags = FOF_NOCONFIRMATION | FOF_SILENT,  // Do not prompt user
+      .fAnyOperationsAborted = FALSE,
+      .lpszProgressTitle = nullptr,
+      .hNameMappings = nullptr
     };
 
     const auto operation_result = SHFileOperation(&file_options);
@@ -123,7 +122,7 @@ public:
 
     // Iterate through all elements in the directory.
     // If the element is a directory, remove all sub-elements recursively.
-    try{
+    try {
       do {
         if (directory_entry_is_clean(directory_entry)) {
           const auto full_file_path{directory_path + "/" + directory_entry->d_name};
--- a/test/temporary_directory_fixture.hpp
+++ b/test/temporary_directory_fixture.hpp
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#ifndef RCPPUTILS__TEMPORARY_DIRECTORY_FIXTURE_HPP_
-#define RCPPUTILS__TEMPORARY_DIRECTORY_FIXTURE_HPP_
+#ifndef TEMPORARY_DIRECTORY_FIXTURE_HPP_
+#define TEMPORARY_DIRECTORY_FIXTURE_HPP_
 
 #include <gmock/gmock.h>
 
@@ -97,4 +97,4 @@ public:
   std::string temporary_dir_path_;
 };
 
-#endif  // RCPPUTILS__TEMPORARY_DIRECTORY_FIXTURE_HPP_
+#endif  // TEMPORARY_DIRECTORY_FIXTURE_HPP_
--- a/test/temporary_directory_fixture.hpp
+++ b/test/temporary_directory_fixture.hpp
@@ -1,0 +1,100 @@
+// Copyright 2018, Bosch Software Innovations GmbH.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RCPPUTILS__TEMPORARY_DIRECTORY_FIXTURE_HPP_
+#define RCPPUTILS__TEMPORARY_DIRECTORY_FIXTURE_HPP_
+
+#include <gmock/gmock.h>
+
+#include <string>
+
+#ifdef _WIN32
+# include <direct.h>
+# include <Windows.h>
+#else
+# include <unistd.h>
+# include <sys/types.h>
+# include <dirent.h>
+#endif
+
+
+class TemporaryDirectoryFixture : public ::testing::Test
+{
+public:
+  TemporaryDirectoryFixture()
+  {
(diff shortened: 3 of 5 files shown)


include/rcpputils/filesystem_helper.hpp
```diff
@@ -217,6 +218,25 @@ inline path remove_extension(const path & file_path, int n_times = 1)
...
+ * structure obtained by stat.
+ * If p does not exist, reports an error.
+ *
+ * \param file_path Path to examine.
+ * \return The size of the file, in bytes.
+ */
+inline std::uintmax_t file_size(const path & file_path)
+{
+  if (!file_path.exists()) {
+    std::error_code ec(ENOENT, std::system_category());
```
zmichaels11:
```c++
const std::error_code ec{ENOENT, std::system_category()};
```
or
```c++
const auto ec = std::error_code{ENOENT, std::system_category()};
```
or
```c++
throw std::system_error({ENOENT, std::system_category()}, "Path does not exist");
```

zmichaels11: Consider including `path` in the error message.

test/test_filesystem_helper.cpp
```diff
@@ -14,16 +14,42 @@
...
+ * @param size File size in MiB
+ */
+void create_file(const std::string & uri, int size)
+{
+  std::ofstream out{uri};
+  if (!out) {
+    throw std::runtime_error("Unable to write file.");
+  }
+  const auto file_text = "test";
+  const auto file_size = size * 1024 * 1024;
```
zmichaels11: define the type of `file_size` here since you're `static_cast<int>` in the following line

test/test_filesystem_helper.cpp
```diff
@@ -188,3 +214,27 @@ TEST(TestFilesystemHelper, remove_extension_no_extension)
   p = rcpputils::fs::remove_extension(p);
   EXPECT_EQ("foo", p.string());
 }
+
+class FilesystemHelperFixture : public TemporaryDirectoryFixture {};
+
+TEST_F(FilesystemHelperFixture, get_file_size)
+{
+  const auto expected_file_size_mib = 1;
```
zmichaels11: use:
```c++
constexpr const std::uintmax_t expected_file_size_mib{1024 * 1024};
```
since you're comparing against a `uintmax_t` later.

test/test_filesystem_helper.cpp
```diff
@@ -188,3 +214,27 @@ TEST(TestFilesystemHelper, remove_extension_no_extension)
...
+
+  const auto file_size = rcpputils::fs::file_size(uri);
+  const std::uintmax_t expected_file_size_bytes = expected_file_size_mib * 1024 * 1024;
+  EXPECT_EQ(file_size, expected_file_size_bytes);
+}
+
+TEST_F(FilesystemHelperFixture, get_file_size_file_does_not_exist)
+{
+  rcpputils::fs::path file_name("file1.txt");
+  const auto uri = path(temporary_dir_path_) / file_name;
```
zmichaels11:
```c++
const rcpputils::fs::path file_name{"file1.txt"};
const auto uri = path{temporary_dir_path_} / file_name;
```

test/test_filesystem_helper.cpp
```diff
@@ -188,3 +214,27 @@ TEST(TestFilesystemHelper, remove_extension_no_extension)
...
   EXPECT_EQ("foo", p.string());
 }
+
+class FilesystemHelperFixture : public TemporaryDirectoryFixture {};
+
+TEST_F(FilesystemHelperFixture, get_file_size)
+{
+  const auto expected_file_size_mib = 1;
+  rcpputils::fs::path file_name("file1.txt");
+  const auto uri = path(temporary_dir_path_) / file_name;
```
zmichaels11:
```c++
const rcpputils::fs::path file_name{"file1.txt"};
const auto uri = path{temporary_dir_path_} / file_name;
```

test/test_filesystem_helper.cpp
```diff
@@ -14,16 +14,42 @@
...
+ * @param uri Path to the file to create.
+ * @param size File size in MiB
+ */
+void create_file(const std::string & uri, int size)
+{
+  std::ofstream out{uri};
+  if (!out) {
+    throw std::runtime_error("Unable to write file.");
+  }
+  const auto file_text = "test";
```
zmichaels11:
```c++
constexpr const char file_text[] = "test";
```

test/test_filesystem_helper.cpp
```diff
@@ -14,16 +14,42 @@
...
 #include <gtest/gtest.h>
 
+#include <fstream>
 #include <string>
 
 #include "rcpputils/filesystem_helper.hpp"
+#include "temporary_directory_fixture.hpp"
 
 #ifdef _WIN32
 static constexpr const bool is_win32 = true;
```
zmichaels11:
```c++
namespace
{
#ifdef _WIN32
constexpr const bool is_win32 = true;
#else
constexpr const bool is_win32 = false;
#endif
}
```

piraka9011: This is something that was already there and not sure if it's something we want to touch in this PR.

zmichaels11: We shouldn't touch it if it was there before.

include/rcpputils/filesystem_helper.hpp
```diff
@@ -229,8 +229,9 @@ inline path remove_extension(const path & file_path, int n_times = 1)
 inline std::uintmax_t file_size(const path & file_path)
 {
   if (!file_path.exists()) {
-    std::error_code ec(ENOENT, std::system_category());
-    throw std::system_error(ec, "Path does not exist");
+    std::error_code ec{ENOENT, std::system_category()};
+    std::string error_msg{"Path " + file_path.string() + " does not exist"};
```
zmichaels11: These can be `const`.

test/test_filesystem_helper.cpp
```diff
@@ -14,16 +14,42 @@
...
+ *
+ * @param uri Path to the file to create.
+ * @param size File size in MiB
+ */
+void create_file(const std::string & uri, int size)
+{
+  std::ofstream out{uri};
+  if (!out) {
+    throw std::runtime_error("Unable to write file.");
+  }
```
zmichaels11: use http://www.cplusplus.com/reference/ios/ios/exceptions/ in place of checking `ofstream` after each operation.

piraka9011: I should've remembered this!

include/rcpputils/filesystem_helper.hpp
```diff
@@ -217,6 +218,26 @@ inline path remove_extension(const path & file_path, int n_times = 1)
...
 
+/**
+ * For a regular file p, returns the size determined as if by reading the st_size member of the
+ * structure obtained by stat.
+ * If p does not exist, reports an error.
+ *
+ * \param file_path Path to examine.
+ * \return The size of the file, in bytes.
+ */
+inline std::uintmax_t file_size(const path & file_path)
```
thomas-moulard: Move implementation to cpp

include/rcpputils/filesystem_helper.hpp
```diff
@@ -217,6 +218,26 @@ inline path remove_extension(const path & file_path, int n_times = 1)
...
+{
+  if (!file_path.exists()) {
+    std::error_code ec{ENOENT, std::system_category()};
+    const std::string error_msg{"Path " + file_path.string() + " does not exist"};
+    throw std::system_error(ec, error_msg);
+  }
+  struct stat stat_buffer {};
```
thomas-moulard:
```suggestion
  struct stat stat_buffer{};
```

include/rcpputils/filesystem_helper.hpp
```diff
@@ -217,6 +218,26 @@ inline path remove_extension(const path & file_path, int n_times = 1)
...
+  int rc = stat(file_path.string().c_str(), &stat_buffer);
```
thomas-moulard:
```suggestion
  const int rc = stat(file_path.string().c_str(), &stat_buffer);
```

nit

include/rcpputils/filesystem_helper.hpp
thomas-moulard: Includes are missing for stat

test/temporary_directory_fixture.hpp
```diff
@@ -0,0 +1,100 @@
...
+# include <dirent.h>
+#endif
+
+
+class TemporaryDirectoryFixture : public ::testing::Test
+{
+public:
+  TemporaryDirectoryFixture()
+  {
+    char template_char[] = "tmp_test_dir.XXXXXX";
```
thomas-moulard: :warning: This code will segfault.

This literal cannot be modified, you should not put it in a const and pass that to function expecting a `char *` as they will modfiy it.

See the longer version here: https://stackoverflow.com/a/31846777/1442517

test/temporary_directory_fixture.hpp
```diff
@@ -0,0 +1,100 @@
...
+    char temp_path[255];
+    GetTempPathA(255, temp_path);
+    _mktemp_s(template_char, strnlen(template_char, 20) + 1);
+    temporary_dir_path_ = std::string(temp_path) + std::string(template_char);
+    _mkdir(temporary_dir_path_.c_str());
+#else
+    char * dir_name = mkdtemp(template_char);
+    temporary_dir_path_ = dir_name;
+#endif
+  }
```
thomas-moulard: This needs some cleanup.

test/temporary_directory_fixture.hpp
```diff
@@ -0,0 +1,100 @@
...
+
+  ~TemporaryDirectoryFixture() override
+  {
+    remove_directory_recursively(temporary_dir_path_);
+  }
+
+  void remove_directory_recursively(const std::string & directory_path)
```
thomas-moulard: Missing an anonymous namespace.

test/temporary_directory_fixture.hpp
```diff
@@ -0,0 +1,100 @@
...
+  {
+#ifdef _WIN32
+    // We need a string of type PCZZTSTR, which is a double null terminated char ptr
+    size_t length = strlen(directory_path.c_str());
+    TCHAR * temp_dir = new TCHAR[length + 2];
+    memcpy(temp_dir, temporary_dir_path_.c_str(), length);
+    temp_dir[length] = 0;
+    temp_dir[length + 1] = 0;  // double null terminated
+
+    SHFILEOPSTRUCT file_options;
```
thomas-moulard:
```c++
    SHFILEOPSTRUCT file_options{
      .hwnd = nullptr,
      .wFunc =...
    };
```

would be preferable, see https://en.cppreference.com/w/c/language/struct_initialization

test/temporary_directory_fixture.hpp
```diff
@@ -0,0 +1,100 @@
...
+  {
+    remove_directory_recursively(temporary_dir_path_);
+  }
+
+  void remove_directory_recursively(const std::string & directory_path)
+  {
+#ifdef _WIN32
+    // We need a string of type PCZZTSTR, which is a double null terminated char ptr
+    size_t length = strlen(directory_path.c_str());
+    TCHAR * temp_dir = new TCHAR[length + 2];
```
thomas-moulard: Use an unique pointer

test/temporary_directory_fixture.hpp
```diff
@@ -0,0 +1,100 @@
...
+# include <unistd.h>
+# include <sys/types.h>
+# include <dirent.h>
+#endif
+
+
+class TemporaryDirectoryFixture : public ::testing::Test
+{
+public:
+  TemporaryDirectoryFixture()
```
thomas-moulard: 1/ this code needs to be re-architectured into small functions, the functions which are generic enough should be moved to rcpputils itself, and should have tests. They should also be moved to their own PRs.
2/ fixtures ctors should not do anything if it can be avoided, those complex operations should be moved to `SetUp` and `TearDown` instead.
3/ This test should be run with `--gtest_repeat` to ensure the creation/destruction logic is working properly in presence of flags.

test/temporary_directory_fixture.hpp
```diff
@@ -0,0 +1,100 @@
...
+    file_options.pTo = nullptr;
+    file_options.fFlags = FOF_NOCONFIRMATION | FOF_SILENT;  // do not prompt user
+    file_options.fAnyOperationsAborted = FALSE;
+    file_options.lpszProgressTitle = nullptr;
+    file_options.hNameMappings = nullptr;
+
+    SHFileOperation(&file_options);
+    delete[] temp_dir;
+#else
+    DIR * dir = opendir(directory_path.c_str());
```
thomas-moulard:
```suggestion
    const DIR * dir = opendir(directory_path.c_str());
```

thomas-moulard: This is not exception safe. We should probably divide that into two functions:

```c++
   void f() {
     const DIR * dir = opendir();
     try {
        g(dir);
     } catch (...) {
        closedir(dir);
        throw;
     }
     closedir(dir);
   }

  void g(DIR * const dir) {
    // ...
  }
```

test/temporary_directory_fixture.hpp
```diff
@@ -0,0 +1,100 @@
...
+      if (strcmp(directory_entry->d_name, ".") != 0 && strcmp(directory_entry->d_name, "..") != 0) {
+        if (directory_entry->d_type == DT_DIR) {
+          remove_directory_recursively(directory_path + "/" + directory_entry->d_name);
+        }
+        remove((directory_path + "/" + directory_entry->d_name).c_str());
+      }
+    }
+    closedir(dir);
+    remove(temporary_dir_path_.c_str());
+#endif
```
thomas-moulard: This needs to be rewritten in C++

test/test_filesystem_helper.cpp
```diff
@@ -14,16 +14,40 @@
...
 
+namespace
+{
+/**
+ * Create a file with a specific size specified in MiB.
+ *
+ * @param uri Path to the file to create.
+ * @param size File size in MiB
+ */
+void create_file(const std::string & uri, int size)
```
thomas-moulard:
```suggestion
void create_file(const std::string & uri, int size_in_mib)
```

test/test_filesystem_helper.cpp
```diff
@@ -14,16 +14,40 @@
...
+{
+  std::ofstream out{uri};
+  out.exceptions(std::ifstream::failbit | std::ifstream::badbit);
+  constexpr const char file_text[] = "test";
+  const auto file_size = size * 1024 * 1024;
```
thomas-moulard:
```suggestion
  const auto size_in_bytes = size_in_mib * 1024 * 1024;
```

test/test_filesystem_helper.cpp
```diff
@@ -14,16 +14,40 @@
...
+ * Create a file with a specific size specified in MiB.
+ *
+ * @param uri Path to the file to create.
+ * @param size File size in MiB
+ */
+void create_file(const std::string & uri, int size)
+{
+  std::ofstream out{uri};
+  out.exceptions(std::ifstream::failbit | std::ifstream::badbit);
+  constexpr const char file_text[] = "test";
```
thomas-moulard:
```suggestion
  const std::string file_text{"test"};
```

test/test_filesystem_helper.cpp
```diff
@@ -14,16 +14,40 @@
...
+  const auto file_size = size * 1024 * 1024;
+  const auto num_iterations = file_size / static_cast<int>(strlen(file_text));
```
thomas-moulard:
```suggestion
  const auto num_iterations = file_size / file_text.size();
```

test/test_filesystem_helper.cpp
```diff
@@ -14,16 +14,40 @@
...
+
+  for (int i = 0; i < num_iterations; i++) {
```
thomas-moulard:
```suggestion
  for (auto i = 0; i < num_iterations; ++i) {
```

piraka9011:
@thomas-moulard FYI, `temporary_directory_fixture.hpp` was taken from `rosbag2_storage` so was unsure how much of it we'd need to revise accordingly.
I'll refactor some of the utility functions and break it down as needed.

@zmichaels11 Take a look at some of the `constexpr`/`char[]` suggestions.

include/rcpputils/filesystem_helper.hpp
piraka9011: Just wondering if there's a reason `remove_extension` was inline but not this.
Is it b/c that one operated on or returned a path?

thomas-moulard:
We should probably clean both IMHO.

test/temporary_directory_fixture.hpp
piraka9011: Interesting, would the better approach be to use a  `std::vector<char>` or a `std::unique_ptr<char[]>` instead?

zmichaels11: I think the correct solution is to strcpy "tmp_test_dir.XXXXXX" into a `char[]` of appropriate size.

So something like:
```c++
const template_char[SOME_MAX_SIZE];  // SOME_MAX_SIZE should be larger than the template string.
strcpy(template_char, "tmp_test_dir.XXXXXX");
```

thomas-moulard: I am wrong. @zmichaels11 no need to specify the string length, this is brittle:
According to https://www.geeksforgeeks.org/char-vs-stdstring-vs-char-c/ and my tests:

```c++
char * foo = "abcd";
foo[0] = 'e'; // segv

char foo2[] = "abcd";
foo2[0] = 'e'; // no segv
```

so it seems safe. @zmichaels11 does that match what you saw?

See `-Wwrite-strings` GCC flag.

thomas-moulard: We could repro with @zmichaels11 - the initial code works as attended and is correct

zmichaels11: Link to reproduction: https://github.com/ros2/rosbag2/pull/239#issuecomment-567690147

test/temporary_directory_fixture.hpp
piraka9011: Can't have this as `const` since used by `readdir()`.

piraka9011:
> Unblocking this I'm OOO from next week.

If you have any comments/suggestions tho lmk!

piraka9011:
After discussion with @emersonknapp , this may not be the best approach, rather, create a wrapper around the C++17 FileSystem API with CMake linkers to C bindings.

CMakeLists.txt
```diff
@@ -24,7 +24,8 @@ include_directories(include)
 ament_export_include_directories(include)
 
 add_library(${PROJECT_NAME}
-  src/find_library.cpp)
+  src/find_library.cpp
+  src/filesystem_helper.cpp)
```
thomas-moulard: alphasort

test/temporary_directory_fixture.hpp
```diff
@@ -0,0 +1,150 @@
...
+ * Create a string of type PCZZTSTR, which is a double null terminated char pointer.
+ *
+ * \param text The string to convert.
+ */
+std::unique_ptr<TCHAR> create_double_null_terminated_tchar(const std::string & text)
+{
+  // Increase length by two to add the double null termination
+  const auto char_length = strlen(text.c_str());
+  const auto new_text = std::make_unique<TCHAR[]>(char_length + 2);
+  // memcpy(temp_dir, text.c_str(), length);
```
thomas-moulard: remove

test/temporary_directory_fixture.hpp
```diff
@@ -0,0 +1,150 @@
...
+
+  void TearDown() override
+  {
+    remove_directory_recursively(temporary_dir_path_);
+  }
+
+  void remove_directory_recursively(const std::string & directory_path)
+  {
+#ifdef _WIN32
+    const auto null_terminated_directory = create_double_null_terminated_tchar(directory_path);
```
thomas-moulard: This is never used?

piraka9011:
Resolved by [ros2/rcutils#197](https://github.com/ros2/rcutils/pull/197)

PR #30, opened 2020-01-14
Fix filesystem helpers for directory manipulation.

hidmic:
Precisely what the title says. Also added a partial `std::filesystem::remove()` equivalent for `rcpputils::fs::create_directories()` testing purposes.

### DIFF
--- a/include/rcpputils/filesystem_helper.hpp
+++ b/include/rcpputils/filesystem_helper.hpp
@@ -182,8 +182,8 @@ inline bool exists(const path & path_to_check)
 inline bool create_directories(const path & p)
 {
   path p_built;
-
   int status = 0;
+
   for (auto it = p.cbegin(); it != p.cend() && status == 0; ++it) {
     if (p_built.empty()) {
       p_built = *it;
--- a/include/rcpputils/filesystem_helper.hpp
+++ b/include/rcpputils/filesystem_helper.hpp
@@ -183,16 +183,31 @@ inline bool create_directories(const path & p)
 {
   path p_built;
 
-  for (auto it = p.cbegin(); it != p.cend(); ++it) {
-    p_built /= *it;
-
+  int status = 0;
+  for (auto it = p.cbegin(); it != p.cend() && status == 0; ++it) {
+    if (p_built.empty()) {
+      p_built = *it;
+    } else {
+      p_built /= *it;
+    }
+    if (!p_built.exists()) {
 #ifdef _WIN32
-    _mkdir(p_built.string().c_str());
+      status = _mkdir(p_built.string().c_str());
 #else
-    mkdir(p_built.string().c_str(), S_IRWXU | S_IRWXG | S_IROTH | S_IXOTH);
+      status = mkdir(p_built.string().c_str(), S_IRWXU | S_IRWXG | S_IROTH | S_IXOTH);
 #endif
+    }
   }
-  return true;
+  return status == 0;
+}
+
+inline bool remove(const path & p)
+{
+#ifdef _WIN32
+  return _rmdir(p.string().c_str()) == 0;
+#else
+  return rmdir(p.string().c_str()) == 0;
+#endif
 }
 
 /**
--- a/include/rcpputils/filesystem_helper.hpp
+++ b/include/rcpputils/filesystem_helper.hpp
@@ -194,7 +194,7 @@ inline bool create_directories(const path & p)
 #ifdef _WIN32
       status = _mkdir(p_built.string().c_str());
 #else
-      status = mkdir(p_built.string().c_str(), S_IRWXU | S_IRWXG | S_IROTH | S_IXOTH);
+      status = mkdir(p_built.string().c_str(), S_IRWXU | S_IRWXG | S_IRWXO);
 #endif
     }
   }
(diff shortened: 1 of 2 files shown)


hidmic:
CI up to `rcpputils`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9022/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4782/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7383/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8956/)

include/rcpputils/filesystem_helper.hpp
```diff
@@ -183,16 +183,31 @@ inline bool create_directories(const path & p)
...
+    }
   }
-  return true;
+  return status == 0;
+}
+
+inline bool remove(const path & p)
+{
+#ifdef _WIN32
+  return _rmdir(p.string().c_str()) == 0;
```
ivanpauno: What if the path is a file?
`remove` is supposed to delete both files and empty directories.
See https://en.cppreference.com/w/cpp/filesystem/remove.

include/rcpputils/filesystem_helper.hpp
```diff
@@ -183,16 +183,31 @@ inline bool create_directories(const path & p)
 {
   path p_built;
```
ivanpauno: nit: delete this new line and add one between `status` and the `for loop`?

include/rcpputils/filesystem_helper.hpp
```diff
@@ -183,16 +183,31 @@ inline bool create_directories(const path & p)
...
 
-  for (auto it = p.cbegin(); it != p.cend(); ++it) {
-    p_built /= *it;
-
+  int status = 0;
+  for (auto it = p.cbegin(); it != p.cend() && status == 0; ++it) {
+    if (p_built.empty()) {
+      p_built = *it;
```
ivanpauno: This just happens in the first iteration, right?
Consider doing it before the loop (though I don't mind much).

include/rcpputils/filesystem_helper.hpp
```diff
@@ -183,16 +183,31 @@ inline bool create_directories(const path & p)
...
+    } else {
+      p_built /= *it;
+    }
+    if (!p_built.exists()) {
 #ifdef _WIN32
-    _mkdir(p_built.string().c_str());
+      status = _mkdir(p_built.string().c_str());
 #else
-    mkdir(p_built.string().c_str(), S_IRWXU | S_IRWXG | S_IROTH | S_IXOTH);
+      status = mkdir(p_built.string().c_str(), S_IRWXU | S_IRWXG | S_IROTH | S_IXOTH);
```
ivanpauno: According to [create_directory docs](https://en.cppreference.com/w/cpp/filesystem/create_directory), the directory is created with `0777` perms (see [here](http://en.cppreference.com/w/cpp/filesystem/perms)).

ivanpauno: Also, shouldn't break the loop if status is not zero?

include/rcpputils/filesystem_helper.hpp
hidmic: Yeap, but then I'm forced to copy loop code. Tried to make it not so, but it gets ugly. Compiler will likely do loop unrolling anyways.

include/rcpputils/filesystem_helper.hpp
hidmic: That's why I said it's a partial implementation. We can make it full, I simply didn't have a use case.

include/rcpputils/filesystem_helper.hpp
hidmic: > According to `create_directory` docs, the directory is created with 0777 perms (see here).

True. I do not know why the original author made it `0775` instead though.

> Also, shouldn't break the loop if status is not zero?

 It is.

include/rcpputils/filesystem_helper.hpp
hidmic: A bit of a taste nit, but sure, either way is fine. See https://github.com/ros2/rcpputils/pull/30/commits/6ac7378a787c17871598150f66de498b9c68131a.

include/rcpputils/filesystem_helper.hpp
ivanpauno: Add a comment in the docsrting of the function. If not, later users might not realize.

include/rcpputils/filesystem_helper.hpp
ivanpauno: > It is.

Didn't realize.

hidmic:
So, @ivanpauno has two good points:

- `rcpputils::fs::create_directories` should use `0777` permissions to fully imitate `std::filesystem::create_directories` -- and to actually be consistent with its behavior in Windows.
- `rcpputils::fs::remove` should also allow file removal to fully imitate its `std::filesystem` counterpart.

@thomas-moulard any comments in favor or against?

jacobperron:
+1 on both raised points in order to match the behavior of `std` equivalents.

hidmic:
@ivanpauno @jacobperron alright, done in https://github.com/ros2/rcpputils/pull/30/commits/449e1a42c8539f9be60443e3d2021d0f61c423f2.

prajakta-gokhale:
Re-running CI for `rcpputils`

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9036/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4791/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7392/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8966/)

Issue #29, opened 2020-01-13
Create a utility function that wraps assert but instead throws


piraka9011:
Inspired by [Kotlin's Assert](https://www.bignerdranch.com/blog/write-better-code-using-kotlins-require-check-and-assert/), it would be helpful to have a wrapper around assert that instead throws to provide a meaningful stack trace.

emersonknapp:
Do you mean a wrapper around assert? Or do you mean an assert-like functionality that works in non-debug builds?

zmichaels11:
I think it would be better as `assert-like`

tfoote:
Please consider picking up https://github.com/ros2/rcutils/pull/112 that got deprioritized previously.

zmichaels11:
Possible implementation: https://github.com/ros2/rcpputils/pull/31

wjwwood:
What is the benefit of having an assert that throws over the standard one that uses abort? Doesn't gdb break on abort just like it does with an exception? Or were you thinking of catching the exceptions? I'm just trying to understand the motivation.

I do understand a `require` or `check` that always asserts even when in release mode.

wjwwood:
> Please consider picking up ros2/rcutils#112 that got deprioritized previously.

This could be a good solution, but note that this implementation is a bit different, as in the case that the condition fails it logs to the rcutils logger (which may or may not be desired) and it uses a "DebugBreak" which is a trap, rather than throwing an exception (which it cannot do since it might be used in C only code) and rather than using `abort()` which on Linux would result in a `SIGABRT` (I think).

If the main goal is to have an `assert()` that works in release mode too (a.k.a. `check` or `required`) then ros2/rcutils#112 might be overkill.

zmichaels11:
A benefit of throwing vs abort is that failure can be checked in unit tests.

wjwwood:
I see, Google Test has "Death tests" which work with asserts but I guess they're hard to use and sometimes don't work as expected on Windows:

https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#how-to-write-a-death-test

zmichaels11:
Also, an exception carries the additional information that the code crashed due to a `check`, `require`, or `assert`

wjwwood:
> Also, an exception carries the additional information that the code crashed due to a check, require, or assert

You mean in gdb? You should be able to see the same kind of information as with an unhandled exception, but I haven't compared the two closely.

[...]

PR #31, opened 2020-01-23
Add functions for C++ assertions

zmichaels11:
### Changes
* Add `rcpputils::require_true(condition)` which throws `std::invalid_argument` if condition is false. Its indended use is for parameter checks that should always be ran.
* Add `rcpputils::check_true(condition)` which throws `rcpputils::InvalidStateException` if condition is false. Its intended use is for state checks that should always be ran.
* Add `rcpputils::assert_true(condition)` which throws `rcpputils::AssertionException` if condition is false AND the macro`NDEBUG` is not set. Its intended use is for additional checks inserted into debug builds.

### Issues
* Feature requested here: https://github.com/ros2/rcpputils/issues/29

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,7 @@ include_directories(include)
 ament_export_include_directories(include)
 
 add_library(${PROJECT_NAME}
+  src/asserts.cpp
   src/find_library.cpp)
 target_include_directories(${PROJECT_NAME}
   PUBLIC
@@ -47,10 +48,12 @@ if(BUILD_TESTING)
   ament_lint_auto_find_test_dependencies()
 
   ament_add_gtest(test_asserts_ndebug test/test_asserts.cpp)
+  target_link_libraries(test_asserts_ndebug ${PROJECT_NAME})
 
   target_compile_definitions(test_asserts_ndebug PUBLIC NDEBUG)
 
   ament_add_gtest(test_asserts_debug test/test_asserts.cpp)
+  target_link_libraries(test_asserts_debug ${PROJECT_NAME})
 
   ament_add_gtest(test_thread_safety_annotations test/test_thread_safety_annotations.cpp)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,9 +48,7 @@ if(BUILD_TESTING)
 
   ament_add_gtest(test_asserts_ndebug test/test_asserts.cpp)
 
-  if(TARGET test_asserts_ndebug)
-    add_definitions(-DNDEBUG)
-  endif()
+  target_compile_definitions(test_asserts_ndebug PUBLIC NDEBUG)
 
   ament_add_gtest(test_asserts_debug test/test_asserts.cpp)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,6 +46,14 @@ if(BUILD_TESTING)
 
   ament_lint_auto_find_test_dependencies()
 
+  ament_add_gtest(test_asserts_ndebug test/test_asserts.cpp)
+
+  if(TARGET test_asserts_ndebug)
+    add_definitions(-DNDEBUG)
+  endif()
+
+  ament_add_gtest(test_asserts_debug test/test_asserts.cpp)
+
   ament_add_gtest(test_thread_safety_annotations test/test_thread_safety_annotations.cpp)
 
   ament_add_gtest(test_join test/test_join.cpp)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,7 +50,9 @@ if(BUILD_TESTING)
   ament_add_gtest(test_asserts_ndebug test/test_asserts.cpp)
   target_link_libraries(test_asserts_ndebug ${PROJECT_NAME})
 
-  target_compile_definitions(test_asserts_ndebug PUBLIC NDEBUG)
+  if(TARGET test_asserts_ndebug)
+    target_compile_definitions(test_asserts_ndebug PUBLIC NDEBUG)
+  endif()
 
   ament_add_gtest(test_asserts_debug test/test_asserts.cpp)
   target_link_libraries(test_asserts_debug ${PROJECT_NAME})
--- a/include/rcpputils/asserts.hpp
+++ b/include/rcpputils/asserts.hpp
@@ -25,7 +25,7 @@ namespace rcpputils
  * \param condition
  * \throw std::runtime_error if the condition is not met.
  */
-inline void check(bool condition)
+inline void check_true(bool condition)
 {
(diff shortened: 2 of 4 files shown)


CMakeLists.txt
```diff
@@ -46,6 +46,14 @@ if(BUILD_TESTING)
 
   ament_lint_auto_find_test_dependencies()
 
+  ament_add_gtest(test_asserts_ndebug test/test_asserts.cpp)
+
+  if(TARGET test_asserts_ndebug)
+    add_definitions(-DNDEBUG)
```
piraka9011: Same for windows?

piraka9011:
This LGTM, main concern is whether this works on windows.

include/rcpputils/asserts.hpp
```diff
@@ -0,0 +1,55 @@
...
+}
+
+/**
+ * Asserts that a condition passes.
+ *
+ * This function behaves similar to assert, except that it throws instead of invoking abort().
+ * \param condition
+ * \throw std::runtime_error if the macro NDEBUG is not set and the condition is not met.
+ */
+inline void ros_assert(bool condition)
```
emersonknapp: Do we want to prefix this instead of just having it be `rcpputils::assert`? I don't have a strong opinion but the `ros_` prefix does not seem to be well-established in ROS2 as a pattern

CMakeLists.txt
zmichaels11: It looks like it is since NDEBUG is defined in the standard library.

include/rcpputils/asserts.hpp
zmichaels11: Addressed in notes at the top. `rcpputils::assert` conflicts with `assert` since its a macro. It might be doable since `std::min` has the same problem.

emersonknapp: Ah - my bad for not reading thoroughly. SGTM

zmichaels11:
@ros2/aws-oncall @prajakta-gokhale @thomas-moulard - please run this CI job
Gist: https://gist.githubusercontent.com/zmichaels11/9af526e5a269ea6e1f4d76afa1379aeb/raw/9f8084b3b84736076bea5cdde99b30576154006a/ros2.repos
Job: ci_launcher

zmichaels11:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9124/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4851/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7453/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9047/)

include/rcpputils/asserts.hpp
wjwwood: Other C++ assert libraries do something like `namespace::assert_true(condition)` or `namespace::assert::is_true(condition)`. Maybe `rcpputils::assert_true()` is better?

zmichaels11: I agree with this. I'll change the function signatures to `assert_true` and `check_true`.

CMakeLists.txt
wjwwood: Maybe https://cmake.org/cmake/help/v3.0/command/target_compile_definitions.html? As written this will affect the whole package. Which might be kind of strange, since you'd expect `NDEBUG` to be missing from `librcpputils` if you built with `-DCMAKE_BUILD_TYPE=Release`, but if this test is also being built that would not be the case any longer. Since the thing being tested is an inlined function in a header only, it should be sufficient to just build the test executable's target with `-DNDEBUG`, right?

zmichaels11:
A few changes, some from discussion with @wjwwood 
* renamed `ros_assert` to `assert_true` and `check` to `check_true`
* use `target_compile_definitions` to specify `NDEBUG` in `test_assert_ndebug`.
* add `AssertionException` and `IllegalStateException` for `assert_true` and `check_true` respectively.
* add `require_true` which throws an `std::invalid_argument`

CMakeLists.txt
```diff
@@ -46,6 +47,14 @@ if(BUILD_TESTING)
 
   ament_lint_auto_find_test_dependencies()
 
+  ament_add_gtest(test_asserts_ndebug test/test_asserts.cpp)
+  target_link_libraries(test_asserts_ndebug ${PROJECT_NAME})
```
wjwwood: I think you still might want to check if `test_asserts_ndebug` is a target before using it, as I do not think `ament_add_gtest` guarantees that it creates a target, in the case that google test is not found:

https://github.com/ament/ament_cmake/blob/83e1b8147e2eda7d805e45c163aea60230a4d160/ament_cmake_gtest/cmake/ament_add_gtest.cmake#L21-L22

zmichaels11:
Running CI with changes:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9130/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4852/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7454/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9048/)

zmichaels11:
Rerun Windows CI
* [Build Status](https://ci.ros2.org/job/ci_windows/9049/)

zmichaels11:
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/9050/)

include/rcpputils/asserts.hpp
```diff
@@ -0,0 +1,103 @@
...
+#include <stdexcept>
+#include <string>
+
+#include "rcpputils/visibility_control.hpp"
+
+// Needed to disable compiler warning for exporting a class that extends a
+// non-DLL-interface class.
+// This should be fine since its extending an STL class.
+#ifdef _WIN32
+# pragma warning(push)
```
wjwwood: Don't you need to pop this at some point?

zmichaels11: yes.

zmichaels11:
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/9052/)

zmichaels11:
Ok I think I got it this time.
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/9053/)

zmichaels11:
Does this require CI to rerun for Linux/macOS?

piraka9011:
> Does this require CI to rerun for Linux/macOS?

Better safe than sorry :)

zmichaels11:
Rest of CI:
* Linux [Build Status](https://ci.ros2.org/job/ci_linux/9132/)
* Linux-aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/4854/)
* macOS (queued)

PR #42, opened 2020-02-07
Make assert functions accept an optional string.

zmichaels11:
### Changes
* Add default msg parameter to `rcpputils::require_true`
* Add default msg parameter to `rcpputils::check_true`
* Add default msg parameter to `rcpputils::assert_true`

Signed-off-by: Zachary Michaels <zmichaels11@gmail.com>

### DIFF
--- a/include/rcpputils/asserts.hpp
+++ b/include/rcpputils/asserts.hpp
@@ -60,7 +60,7 @@ public:
  * \param msg
  * \throw std::invalid_argument if the condition is not met.
  */
-inline void require_true(bool condition, const std::string & msg = "Invalid argument passed!")
+inline void require_true(bool condition, const std::string & msg = "invalid argument passed")
 {
   if (!condition) {
     throw std::invalid_argument{msg};
@@ -74,7 +74,7 @@ inline void require_true(bool condition, const std::string & msg = "Invalid argu
  * \param msg
  * \throw rcpputils::IllegalStateException if the condition is not met.
  */
-inline void check_true(bool condition, const std::string & msg = "Check reported invalid state!")
+inline void check_true(bool condition, const std::string & msg = "check reported invalid state")
 {
   if (!condition) {
     throw rcpputils::IllegalStateException{msg.c_str()};
@@ -89,7 +89,7 @@ inline void check_true(bool condition, const std::string & msg = "Check reported
  * \param msg
  * \throw rcpputils::AssertionException if the macro NDEBUG is not set and the condition is not met.
  */
-inline void assert_true(bool condition, const std::string & msg = "Assertion failed!")
+inline void assert_true(bool condition, const std::string & msg = "assertion failed")
 {
 // Same macro definition used by cassert
 #ifndef NDEBUG
--- a/include/rcpputils/asserts.hpp
+++ b/include/rcpputils/asserts.hpp
@@ -63,7 +63,7 @@ public:
 inline void require_true(bool condition, const std::string & msg = "Invalid argument passed!")
 {
   if (!condition) {
-    throw std::invalid_argument{msg.c_str()};
+    throw std::invalid_argument{msg};
   }
 }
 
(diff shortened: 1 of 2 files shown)


piraka9011:
Is this a requested feature and does it make sense to allow someone to provide custom strings for the error message?
Wouldn't it be easier if the message was standardized so that if someone was using this, they have a clear understanding of where the issue is coming from?

zmichaels11:
@piraka9011 I wanted to replace some state checks with `rcpputils::check_true` now that `rcpputils/asserts.hpp` is available on master and I didn't want to lose the message tied to the old state checks.

For context, the code looked something like:
```c++
void some_function()
{
  if (rcpputils::fs::exists("some path")) {
    throw std::runtime_error{"some path must exist!"};
  }
  logic();
}
```
and replace it with:
```c++
void some_function()
{
  rcpputils::check_true(rcpputils::fs::exists("some path"), "some path must exist!");
  logic();
}
```

include/rcpputils/asserts.hpp
```diff
@@ -57,25 +57,27 @@ class RCPPUTILS_PUBLIC IllegalStateException : public std::exception
...
  * \param condition
+ * \param msg
  * \throw std::invalid_argument if the condition is not met.
  */
-inline void require_true(bool condition)
+inline void require_true(bool condition, const std::string & msg = "Invalid argument passed!")
 {
   if (!condition) {
-    throw std::invalid_argument{"Invalid argument passed!"};
+    throw std::invalid_argument{msg.c_str()};
```
mm318: This can just take the `std::string` directly, that is don't do `.cstr()`.

mm318: Please check the other cases as well.

zmichaels11: Only `std::invalid_argument` accepts an `std::string`.

include/rcpputils/asserts.hpp
```diff
@@ -57,25 +57,27 @@ class RCPPUTILS_PUBLIC IllegalStateException : public std::exception
  * Checks that an argument condition passes.
  *
  * \param condition
+ * \param msg
  * \throw std::invalid_argument if the condition is not met.
  */
-inline void require_true(bool condition)
+inline void require_true(bool condition, const std::string & msg = "Invalid argument passed!")
```
wjwwood: I didn't bring this up on the original pull request, but I keep seeing it, so I'll mention it now. We should probably avoid using `!` in error messages. It's an exception or assertion failure, we already know its important, so I don't think it needs extra punctuation.

I'd recommend following the examples like the STL or glibc, which both use lowercase and no punctuation everywhere. This allows others to combine your error message into a larger string more easily without it looking strange.

For example, consider:

```c++
require_true(false);
```

Depending on how it's handled it might look like this:

```
unhandled exception: invalid argument passed
```

Or:

```
[WARN]: Failed to do X due to invalid argument passed, retrying...
```

If you have punctuation and/or capitalization, it ends up:

```
[WARN]: Failed to do X due to Invalid argument passed!, retrying...
```

It's a really minor thing, but I think it ends up looking more professional to avoid things like exclamation marks, at least in the default messages. If you choose to use them at the user level, that may be totally fine.

zmichaels11:
@ros2/aws-oncall - please run this CI job
Gist: https://gist.githubusercontent.com/zmichaels11/be02caa4917e0bb0a0775ed34fd1d7cd/raw/f106fa8933157e6535d3f336c81e8bf281aeeff3/ros2.repos
Job: ci_launcher

zmichaels11:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9272/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4959/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7586/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9193/)
* Windows-container [Build Status](http://ci.ros2.org/job/ci_windows-container/77/)

PR #44, opened 2020-02-12
Search RUNPATH to find library on Linux

### LINKED ISSUE #40: find_library_path() for rmw fails when running node with root capabilities
I already mentioned this in https://github.com/ros2/rcutils/issues/143, but thought it would warrant its own issue, as the use case might be common.

If a binary is run with capabilities set via `setcap` or with the setuid bit, `LD_LIBRARY_PATH` is omitted during execution. Since `find_library_path()` relies exclusively on `LD_LIBRARY_PATH`, a ros2 node requiring root capabilities fails with

```
terminate called after throwing an instance of 'rclcpp::exceptions::RCLError'
  what():  failed to initialized rcl init options: failed to find shared library of rmw implementation. Searched rmw_fastrtps_cpp, at /tmp/binarydeb/ros-eloquent-rmw-implementation-0.8.2/src/functions.cpp:130, at /tmp/binarydeb/ros-eloquent-rcl-0.8.3/src/rcl/init_options.c:55
```

In our case, we need raw socket access in our node. Maybe introducing another environment variable like `RMW_LIBRARY_PATH` could solve this, or reading out `RPATH` from the binary header. Any suggestions for a workaround in the meantime would be greatly appreciated!

wieset:
Proposal to fix #40

Based on a code snippet from Stack Overflow:
https://stackoverflow.com/questions/2836330/is-there-a-way-to-inspect-the-current-rpath-on-linux

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,6 +34,10 @@ if(WIN32)
   target_compile_definitions(${PROJECT_NAME}
     PRIVATE "RCPPUTILS_BUILDING_LIBRARY")
 endif()
+set_target_properties(${PROJECT_NAME} PROPERTIES
+  BUILD_RPATH ${CMAKE_CURRENT_BINARY_DIR}
+  INSTALL_RPATH_USE_LINK_PATH TRUE
+)
 ament_target_dependencies(${PROJECT_NAME} rcutils)
 ament_export_libraries(${PROJECT_NAME})
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -75,12 +75,31 @@ if(BUILD_TESTING)
 
   ament_add_gtest(test_endian test/test_endian.cpp)
 
+  if(WIN32)
+    set(LIBRARY_PATH_VAR "PATH")
+  elseif(APPLE)
+    set(LIBRARY_PATH_VAR "DYLD_LIBRARY_PATH")
+  else()
+    set(LIBRARY_PATH_VAR "LD_LIBRARY_PATH")
+  endif()
+
   add_library(test_library SHARED test/test_library.cpp)
+
   ament_add_gtest(test_find_library test/test_find_library.cpp)
   target_link_libraries(test_find_library ${PROJECT_NAME} test_library)
   set_tests_properties(test_find_library PROPERTIES
     ENVIRONMENT
-      "_TEST_LIBRARY_DIR=$<TARGET_FILE_DIR:test_library>;_TEST_LIBRARY=$<TARGET_FILE:test_library>")
+      "${LIBRARY_PATH_VAR}=$<TARGET_FILE_DIR:test_library>;_TEST_LIBRARY=$<TARGET_FILE:test_library>")
+
+  # Test with empty LD_LIBRARY_PATH to emulate setcap / setuid executable.
+  # Using RUNPATH to find library instead.
+  if(NOT (WIN32 OR APPLE))
+    ament_add_gtest(test_find_library_runpath test/test_find_library.cpp)
+    target_link_libraries(test_find_library_runpath ${PROJECT_NAME} test_library)
+    set_tests_properties(test_find_library_runpath PROPERTIES
+      ENVIRONMENT
+        "${LIBRARY_PATH_VAR}=;_TEST_LIBRARY=$<TARGET_FILE:test_library>")
+  endif()
 endif()
 
 ament_package()
--- a/src/find_library.cpp
+++ b/src/find_library.cpp
@@ -21,6 +21,7 @@
 #include <fstream>
 #include <sstream>
 #include <string>
+#include <link.h>
 
 #include "rcutils/filesystem.h"
 #include "rcutils/get_env.h"
@@ -75,6 +76,27 @@ std::string find_library_path(const std::string & library_name)
 {
   std::string search_path = get_env_var(kPathVar);
   std::list<std::string> search_paths = split(search_path, kPathSeparator);
+
+#if !defined(_WIN32) && !defined(__APPLE__)
+  std::string search_path_runpath;
+
+  const ElfW(Dyn) *dyn = _DYNAMIC;
+  const ElfW(Dyn) *runpath = NULL;
+  const char *strtab = NULL;
+
+  for (; dyn->d_tag != DT_NULL; ++dyn) {
+    if (dyn->d_tag == DT_RUNPATH) {
+      runpath = dyn;
+    } else if (dyn->d_tag == DT_STRTAB) {
+      strtab = (const char *)dyn->d_un.d_val;
+    }
+  }
+
+  if (strtab != NULL && runpath != NULL) {
+    search_path_runpath = std::string(strtab + runpath->d_un.d_val);
+    search_paths.splice(search_paths.cend(), split(search_path_runpath, kPathSeparator));
+  }
+#endif
 
   std::string filename = kSolibPrefix;
   filename += library_name + kSolibExtension;
--- a/src/find_library.cpp
+++ b/src/find_library.cpp
@@ -81,9 +81,9 @@ std::string find_library_path(const std::string & library_name)
 #if !defined(_WIN32) && !defined(__APPLE__)
   std::string search_path_runpath;
 
-  const ElfW(Dyn) *dyn = _DYNAMIC;
-  const ElfW(Dyn) *runpath = NULL;
-  const char *strtab = NULL;
+  const ElfW(Dyn) * dyn = _DYNAMIC;
+  const ElfW(Dyn) * runpath = NULL;
+  const char * strtab = NULL;
 
   for (; dyn->d_tag != DT_NULL; ++dyn) {
     if (dyn->d_tag == DT_RUNPATH) {
--- a/src/find_library.cpp
+++ b/src/find_library.cpp
@@ -106,7 +106,7 @@ std::list<std::string> retrieve_runpath_paths()
 
 std::string find_library_path(const std::string & library_name)
 {
-  std::string search_path_env = get_env_var(kPathVar);
+  const std::string search_path_env = get_env_var(kPathVar);
   std::list<std::string> search_paths = split(search_path_env, kPathSeparator);
 
   std::list<std::string> search_paths_runpath = retrieve_runpath_paths();
--- a/src/find_library.cpp
+++ b/src/find_library.cpp
@@ -14,7 +14,10 @@
 
 #include "rcpputils/find_library.hpp"
 
+
+#if !defined(_WIN32) && !defined(__APPLE__)
 #include <link.h>
+#endif
 
 #include <cassert>
 #include <cstddef>
--- a/src/find_library.cpp
+++ b/src/find_library.cpp
@@ -74,16 +74,11 @@ std::list<std::string> split(const std::string & value, const char delimiter)
   return list;
 }
 
-}  // namespace
-
-std::string find_library_path(const std::string & library_name)
-{
-  std::string search_path = get_env_var(kPathVar);
-  std::list<std::string> search_paths = split(search_path, kPathSeparator);
-
 #if !defined(_WIN32) && !defined(__APPLE__)
-  std::string search_path_runpath;
-
+// Retrieves a list of paths from the RUNPATH header on Linux.
+// Useful when LD_LIBRARY_PATH is ignored for setcap / setuid executables.
+std::list<std::string> retrieve_runpath_paths()
+{
   const ElfW(Dyn) * dyn = _DYNAMIC;
   const ElfW(Dyn) * runpath = NULL;
   const char * strtab = NULL;
@@ -96,10 +91,25 @@ std::string find_library_path(const std::string & library_name)
     }
   }
 
+  std::string search_path;
   if (strtab != NULL && runpath != NULL) {
-    search_path_runpath = std::string(strtab + runpath->d_un.d_val);
-    search_paths.splice(search_paths.cend(), split(search_path_runpath, kPathSeparator));
+    search_path = std::string(strtab + runpath->d_un.d_val);
   }
+
+  return split(search_path, kPathSeparator);
+}
+#endif
+
+}  // namespace
+
+std::string find_library_path(const std::string & library_name)
+{
+  std::string search_path = get_env_var(kPathVar);
+  std::list<std::string> search_paths = split(search_path, kPathSeparator);
+
+#if !defined(_WIN32) && !defined(__APPLE__)
+  std::list<std::string> search_paths_runpath = retrieve_runpath_paths();
+  search_paths.splice(search_paths.cend(), search_paths_runpath);
 #endif
 
   std::string filename = kSolibPrefix;
--- a/src/find_library.cpp
+++ b/src/find_library.cpp
@@ -14,6 +14,8 @@
 
 #include "rcpputils/find_library.hpp"
 
+#include <link.h>
+
 #include <cassert>
 #include <cstddef>
 
@@ -21,7 +23,6 @@
 #include <fstream>
 #include <sstream>
 #include <string>
-#include <link.h>
 
 #include "rcutils/filesystem.h"
 #include "rcutils/get_env.h"
--- a/src/find_library.cpp
+++ b/src/find_library.cpp
@@ -14,7 +14,6 @@
 
 #include "rcpputils/find_library.hpp"
 
-
 #if !defined(_WIN32) && !defined(__APPLE__)
 #include <link.h>
 #endif
@@ -74,11 +73,14 @@ std::list<std::string> split(const std::string & value, const char delimiter)
   return list;
 }
 
-#if !defined(_WIN32) && !defined(__APPLE__)
 // Retrieves a list of paths from the RUNPATH header on Linux.
 // Useful when LD_LIBRARY_PATH is ignored for setcap / setuid executables.
 std::list<std::string> retrieve_runpath_paths()
 {
+#if defined(_WIN32) || defined(__APPLE__)
+  // Return empty list for win / macos
+  return std::list<std::string>();
+#else
   const ElfW(Dyn) * dyn = _DYNAMIC;
   const ElfW(Dyn) * runpath = NULL;
   const char * strtab = NULL;
@@ -97,20 +99,18 @@ std::list<std::string> retrieve_runpath_paths()
   }
 
   return split(search_path, kPathSeparator);
-}
 #endif
+}
 
 }  // namespace
 
 std::string find_library_path(const std::string & library_name)
 {
-  std::string search_path = get_env_var(kPathVar);
-  std::list<std::string> search_paths = split(search_path, kPathSeparator);
+  std::string search_path_env = get_env_var(kPathVar);
+  std::list<std::string> search_paths = split(search_path_env, kPathSeparator);
 
-#if !defined(_WIN32) && !defined(__APPLE__)
   std::list<std::string> search_paths_runpath = retrieve_runpath_paths();
   search_paths.splice(search_paths.cend(), search_paths_runpath);
-#endif
 
   std::string filename = kSolibPrefix;
   filename += library_name + kSolibExtension;
--- a/test/test_find_library.cpp
+++ b/test/test_find_library.cpp
@@ -59,8 +59,17 @@ TEST(test_find_library, find_library)
 #endif
 
   // Positive test.
-  const std::string test_lib_actual = find_library_path("test_library");
+  std::string test_lib_actual = find_library_path("test_library");
   EXPECT_EQ(test_lib_actual, expected_library_path);
+
+#if !defined(_WIN32) && !defined(__APPLE__)
+  EXPECT_EQ(setenv(env_var, "", override), 0);
+
+  // Test with empty LD_LIBRARY_PATH to emulate setcap / setuid executable.
+  // Using RUNPATH to find library instead.
+  test_lib_actual = find_library_path("test_library");
+  EXPECT_EQ(test_lib_actual, expected_library_path);
+#endif
 
   // (Hopefully) Negative test.
   const std::string bad_path = find_library_path(
--- a/test/test_find_library.cpp
+++ b/test/test_find_library.cpp
@@ -38,7 +38,7 @@ TEST(test_find_library, find_library)
   }
 
   // Positive test.
-  std::string test_lib_actual = find_library_path("test_library");
+  const std::string test_lib_actual = find_library_path("test_library");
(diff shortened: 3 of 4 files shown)


src/find_library.cpp
```diff
@@ -14,6 +14,8 @@
 
 #include "rcpputils/find_library.hpp"
 
+#include <link.h>
```
thomas-moulard: You'll need to add an ifdef around this as this does not exist on Apple/Windows.

src/find_library.cpp
```diff
@@ -76,6 +78,27 @@ std::string find_library_path(const std::string & library_name)
   std::string search_path = get_env_var(kPathVar);
   std::list<std::string> search_paths = split(search_path, kPathSeparator);
 
+#if !defined(_WIN32) && !defined(__APPLE__)
```
thomas-moulard: Could you extract this logic into a helper method?

```c++
#if !defined(_WIN32) && !defined(__APPLE__)

// <-- explain what the function is doing, and why we're doing this here.
std::list<std::string> retrieve_rpath_paths() {
  std::list<std::string> search_paths;
  std::string search_path_runpath;

  const ElfW(Dyn) * dyn = _DYNAMIC;
  const ElfW(Dyn) * runpath = NULL;
  const char * strtab = NULL;

  for (; dyn->d_tag != DT_NULL; ++dyn) {
    if (dyn->d_tag == DT_RUNPATH) {
      runpath = dyn;
    } else if (dyn->d_tag == DT_STRTAB) {
      strtab = (const char *)dyn->d_un.d_val;
    }
  }

  if (strtab != NULL && runpath != NULL) {
    search_path_runpath = std::string(strtab + runpath->d_un.d_val);
    search_paths.splice(search_paths.cend(), split(search_path_runpath, kPathSeparator));
  }
  return search_paths;
}
#else
std::list<std::string> retrieve_rpath_paths() {
  return std::list<std::string>();
}
#endif
```

...and then you could call the function where the implementation is currently located.

test/test_find_library.cpp
```diff
@@ -59,9 +59,18 @@ TEST(test_find_library, find_library)
...
 
+#if !defined(_WIN32) && !defined(__APPLE__)
+  EXPECT_EQ(setenv(env_var, "", override), 0);
+
+  // Test with empty LD_LIBRARY_PATH to emulate setcap / setuid executable.
+  // Using RUNPATH to find library instead.
+  test_lib_actual = find_library_path("test_library");
+  EXPECT_EQ(test_lib_actual, expected_library_path);
+#endif
+
```
thomas-moulard: Could we move this into a different test case, and use the gtest fixture to clear out and re-enable the `LD_LIBRARY_PATH` value? This does not re-enable the environment variable, so all tests after this one will actually be running in a modified environment. It will also potentially break people relying on gtest flags such as `--gtest_repeat=10` (first one will have full env, next ones will run w/o `LD_LIBRARY_PATH`).

thomas-moulard:
Thanks for your contribution! The approach LGTM, a few nits about the implementation and test.

src/find_library.cpp
wieset: Sure, will fix

src/find_library.cpp
wieset: Ok will do. If you don't mind I'd call it `retrieve_runpath_paths()` since I'm reading out `RUNPATH` as `RPATH` is deprecated.

test/test_find_library.cpp
wieset: How about moving the setting of the env var to the test properties in CMakeLists and reusing the test function for both cases? This would also be more in line with what's actually happening (`LD_LIBRARY_PATH` being dropped at execution). Like this:

```cmake
  if(WIN32)
    SET(LIBRARY_PATH_VAR "PATH")
  elseif(APPLE)
    SET(LIBRARY_PATH_VAR "DYLD_LIBRARY_PATH")
  else()
    SET(LIBRARY_PATH_VAR "LD_LIBRARY_PATH")
  endif()

  add_library(test_library SHARED test/test_library.cpp)

  ament_add_gtest(test_find_library test/test_find_library.cpp)
  target_link_libraries(test_find_library ${PROJECT_NAME} test_library)
  set_tests_properties(test_find_library PROPERTIES
    ENVIRONMENT
      "${LIBRARY_PATH_VAR}=$<TARGET_FILE_DIR:test_library>;_TEST_LIBRARY=$<TARGET_FILE:test_library>")

  if(NOT (WIN32 OR APPLE))
    ament_add_gtest(test_find_library_runpath test/test_find_library.cpp)
    target_link_libraries(test_find_library_runpath ${PROJECT_NAME} test_library)
    set_tests_properties(test_find_library_runpath PROPERTIES
      ENVIRONMENT
        "${LIBRARY_PATH_VAR}=;_TEST_LIBRARY=$<TARGET_FILE:test_library>")
  endif()
```

src/find_library.cpp
```diff
@@ -69,13 +74,44 @@ std::list<std::string> split(const std::string & value, const char delimiter)
...
 
 std::string find_library_path(const std::string & library_name)
 {
   std::string search_path = get_env_var(kPathVar);
   std::list<std::string> search_paths = split(search_path, kPathSeparator);
 
+#if !defined(_WIN32) && !defined(__APPLE__)
+  std::list<std::string> search_paths_runpath = retrieve_runpath_paths();
+  search_paths.splice(search_paths.cend(), search_paths_runpath);
+#endif
```
thomas-moulard: Let's try to minimize the number of ifdefs and have identical symbols throughout all platforms. Do you mind defining returning an empty `retrieve_runpath_paths` on non-Linux platforms?

test/test_find_library.cpp
thomas-moulard: This will still have a global effect and impact other test cases, so this does not really solve the problem at hand IMHO.

wieset: Hmm, on https://cmake.org/cmake/help/latest/prop_test/ENVIRONMENT.html it says
> The environment is restored to its previous state after the test is done.

The env setting code in `test_find_library.cpp` would be removed as this would be handled through the test property in CMake that sets (and restores) the `LIBRARY_PATH_VAR` directly.

src/find_library.cpp
wieset: Sounds good, will do

test/test_find_library.cpp
wieset: Ok, I pushed this change for review and further discussion.

test/test_find_library.cpp
```diff
@@ -37,29 +37,8 @@ TEST(test_find_library, find_library)
...
-#ifdef _WIN32
-  EXPECT_EQ(_putenv_s(env_var, test_lib_dir), 0);
-#else
-  const int override = 1;
-  EXPECT_EQ(setenv(env_var, test_lib_dir, override), 0);
-#endif
-
   // Positive test.
-  const std::string test_lib_actual = find_library_path("test_library");
+  std::string test_lib_actual = find_library_path("test_library");
```
thomas-moulard: revert?

src/find_library.cpp
```diff
@@ -69,12 +73,44 @@ std::list<std::string> split(const std::string & value, const char delimiter)
...
+}
+
 }  // namespace
 
 std::string find_library_path(const std::string & library_name)
 {
-  std::string search_path = get_env_var(kPathVar);
-  std::list<std::string> search_paths = split(search_path, kPathSeparator);
+  std::string search_path_env = get_env_var(kPathVar);
```
thomas-moulard:
```suggestion
  const std::string search_path_env = get_env_var(kPathVar);
```

src/find_library.cpp
```diff
@@ -69,12 +73,44 @@ std::list<std::string> split(const std::string & value, const char delimiter)
...
+  std::list<std::string> search_paths = split(search_path_env, kPathSeparator);
```
thomas-moulard:
```suggestion
  const std::list<std::string> search_paths = split(search_path_env, kPathSeparator);
```

src/find_library.cpp
```diff
@@ -69,12 +73,44 @@ std::list<std::string> split(const std::string & value, const char delimiter)
...
+
+  std::list<std::string> search_paths_runpath = retrieve_runpath_paths();
```
thomas-moulard:
```suggestion
  const std::list<std::string> search_paths_runpath = retrieve_runpath_paths();
```

test/test_find_library.cpp
wieset: Yep, dropped the const by mistake

thomas-moulard:
```
2020-03-16T15:29:09.7674334Z Scanning dependencies of target rcpputils
2020-03-16T15:29:09.7674432Z [ 12%] Building CXX object CMakeFiles/rcpputils.dir/src/asserts.cpp.o
2020-03-16T15:29:09.7674538Z [ 16%] Building CXX object CMakeFiles/rcpputils.dir/src/find_library.cpp.o
2020-03-16T15:29:09.7675520Z /home/runner/work/rcpputils/rcpputils/ros_ws/src/rcpputils/src/find_library.cpp: In function â€˜std::__cxx11::string rcpputils::find_library_path(const string&)â€™:
2020-03-16T15:29:09.7676045Z /home/runner/work/rcpputils/rcpputils/ros_ws/src/rcpputils/src/find_library.cpp:113:64: error: passing â€˜const std::__cxx11::list<std::__cxx11::basic_string<char> >â€™ as â€˜thisâ€™ argument discards qualifiers [-fpermissive]
2020-03-16T15:29:09.7676169Z    search_paths.splice(search_paths.cend(), search_paths_runpath);
2020-03-16T15:29:09.7676276Z                                                                 ^
2020-03-16T15:29:09.7676377Z In file included from /usr/include/c++/7/list:63:0,
2020-03-16T15:29:09.7676485Z                  from /home/runner/work/rcpputils/rcpputils/ros_ws/src/rcpputils/src/find_library.cpp:24:
2020-03-16T15:29:09.7677142Z /usr/include/c++/7/bits/stl_list.h:1443:7: note:   in call to â€˜void std::__cxx11::list<_Tp, _Alloc>::splice(std::__cxx11::list<_Tp, _Alloc>::const_iterator, std::__cxx11::list<_Tp, _Alloc>&) [with _Tp = std::__cxx11::basic_string<char>; _Alloc = std::allocator<std::__cxx11::basic_string<char> >; std::__cxx11::list<_Tp, _Alloc>::const_iterator = std::_List_const_iterator<std::__cxx11::basic_string<char> >]â€™
2020-03-16T15:29:09.7677401Z        splice(const_iterator __position, list& __x) noexcept
2020-03-16T15:29:09.7677499Z        ^~~~~~
2020-03-16T15:29:09.7678051Z /home/runner/work/rcpputils/rcpputils/ros_ws/src/rcpputils/src/find_library.cpp:113:64: error: binding reference of type â€˜std::__cxx11::list<std::__cxx11::basic_string<char> >&â€™ to â€˜const std::__cxx11::list<std::__cxx11::basic_string<char> >â€™ discards qualifiers
2020-03-16T15:29:09.7678184Z    search_paths.splice(search_paths.cend(), search_paths_runpath);
2020-03-16T15:29:09.7678285Z                                                                 ^
2020-03-16T15:29:09.7678440Z In file included from /usr/include/c++/7/list:63:0,
2020-03-16T15:29:09.7678549Z                  from /home/runner/work/rcpputils/rcpputils/ros_ws/src/rcpputils/src/find_library.cpp:24:
2020-03-16T15:29:09.7679199Z /usr/include/c++/7/bits/stl_list.h:1443:7: note:   initializing argument 2 of â€˜void std::__cxx11::list<_Tp, _Alloc>::splice(std::__cxx11::list<_Tp, _Alloc>::const_iterator, std::__cxx11::list<_Tp, _Alloc>&) [with _Tp = std::__cxx11::basic_string<char>; _Alloc = std::allocator<std::__cxx11::basic_string<char> >; std::__cxx11::list<_Tp, _Alloc>::const_iterator = std::_List_const_iterator<std::__cxx11::basic_string<char> >]â€™
2020-03-16T15:29:09.7679345Z        splice(const_iterator __position, list& __x) noexcept
2020-03-16T15:29:09.7679442Z        ^~~~~~
2020-03-16T15:29:09.7679698Z CMakeFiles/rcpputils.dir/build.make:75: recipe for target 'CMakeFiles/rcpputils.dir/src/find_library.cpp.o' failed
2020-03-16T15:29:09.7679815Z make[2]: *** [CMakeFiles/rcpputils.dir/src/find_library.cpp.o] Error 1
2020-03-16T15:29:09.7680059Z CMakeFiles/Makefile2:405: recipe for target 'CMakeFiles/rcpputils.dir/all' failed
2020-03-16T15:29:09.7680167Z make[1]: *** [CMakeFiles/rcpputils.dir/all] Error 2
2020-03-16T15:29:09.7680268Z make[1]: *** Waiting for unfinished jobs....
2020-03-16T15:29:09.7680348Z [ 19%] Linking CXX static library libgtest_main.a
2020-03-16T15:29:09.7680440Z [ 19%] Built target gtest_main
2020-03-16T15:29:09.7680639Z Makefile:140: recipe for target 'all' failed
2020-03-16T15:29:09.7680733Z make: *** [all] Error 2
2020-03-16T15:29:09.7680903Z ---
2020-03-16T15:29:09.7681076Z --- stderr: rcpputils
2020-03-16T15:29:09.7681499Z /home/runner/work/rcpputils/rcpputils/ros_ws/src/rcpputils/src/find_library.cpp: In function â€˜std::__cxx11::string rcpputils::find_library_path(const string&)â€™:
2020-03-16T15:29:09.7681995Z /home/runner/work/rcpputils/rcpputils/ros_ws/src/rcpputils/src/find_library.cpp:113:64: error: passing â€˜const std::__cxx11::list<std::__cxx11::basic_string<char> >â€™ as â€˜thisâ€™ argument discards qualifiers [-fpermissive]
2020-03-16T15:29:09.7682121Z    search_paths.splice(search_paths.cend(), search_paths_runpath);
2020-03-16T15:29:09.7682223Z                                                                 ^
2020-03-16T15:29:09.7682328Z In file included from /usr/include/c++/7/list:63:0,
2020-03-16T15:29:09.7682433Z                  from /home/runner/work/rcpputils/rcpputils/ros_ws/src/rcpputils/src/find_library.cpp:24:
2020-03-16T15:29:09.7683220Z /usr/include/c++/7/bits/stl_list.h:1443:7: note:   in call to â€˜void std::__cxx11::list<_Tp, _Alloc>::splice(std::__cxx11::list<_Tp, _Alloc>::const_iterator, std::__cxx11::list<_Tp, _Alloc>&) [with _Tp = std::__cxx11::basic_string<char>; _Alloc = std::allocator<std::__cxx11::basic_string<char> >; std::__cxx11::list<_Tp, _Alloc>::const_iterator = std::_List_const_iterator<std::__cxx11::basic_string<char> >]â€™
2020-03-16T15:29:09.7683380Z        splice(const_iterator __position, list& __x) noexcept
2020-03-16T15:29:09.7683488Z        ^~~~~~
2020-03-16T15:29:09.7684088Z /home/runner/work/rcpputils/rcpputils/ros_ws/src/rcpputils/src/find_library.cpp:113:64: error: binding reference of type â€˜std::__cxx11::list<std::__cxx11::basic_string<char> >&â€™ to â€˜const std::__cxx11::list<std::__cxx11::basic_string<char> >â€™ discards qualifiers
2020-03-16T15:29:09.7684284Z    search_paths.splice(search_paths.cend(), search_paths_runpath);
2020-03-16T15:29:09.7684397Z                                                                 ^
2020-03-16T15:29:09.7684509Z In file included from /usr/include/c++/7/list:63:0,
2020-03-16T15:29:09.7684627Z                  from /home/runner/work/rcpputils/rcpputils/ros_ws/src/rcpputils/src/find_library.cpp:24:
2020-03-16T15:29:09.7685398Z /usr/include/c++/7/bits/stl_list.h:1443:7: note:   initializing argument 2 of â€˜void std::__cxx11::list<_Tp, _Alloc>::splice(std::__cxx11::list<_Tp, _Alloc>::const_iterator, std::__cxx11::list<_Tp, _Alloc>&) [with _Tp = std::__cxx11::basic_string<char>; _Alloc = std::allocator<std::__cxx11::basic_string<char> >; std::__cxx11::list<_Tp, _Alloc>::const_iterator = std::_List_const_iterator<std::__cxx11::basic_string<char> >]â€™
2020-03-16T15:29:09.7685565Z        splice(const_iterator __position, list& __x) noexcept
2020-03-16T15:29:09.7685679Z        ^~~~~~
2020-03-16T15:29:09.7685786Z make[2]: *** [CMakeFiles/rcpputils.dir/src/find_library.cpp.o] Error 1
2020-03-16T15:29:09.7685902Z make[1]: *** [CMakeFiles/rcpputils.dir/all] Error 2
2020-03-16T15:29:09.7686012Z make[1]: *** Waiting for unfinished jobs....
2020-03-16T15:29:09.7686119Z make: *** [all] Error 2
2020-03-16T15:29:09.7686319Z ---
2020-03-16T15:29:09.7686407Z Failed   <<< rcpputils	[ Exited with code 2 ]
2020-03-16T15:29:09.8171310Z 
2020-03-16T15:29:09.8171894Z Summary: 50 packages finished [59.0s]
2020-03-16T15:29:09.8172064Z   1 package failed: rcpputils
2020-03-16T15:29:09.8172169Z   1 package had stderr output: rcpputils
2020-03-16T15:29:09.8631007Z ##[endgroup]
2020-03-16T15:29:09.8635441Z ##[error]The process 'bash' failed with exit code 2
2020-03-16T15:29:09.8753653Z Cleaning up orphan processes
```

Seems like the CI failed building the code.

src/find_library.cpp
wieset: `search_paths` gets updated in line 113, so cannot be const

src/find_library.cpp
wieset: `splice()` in line 113 manipulates the list, so cannot be const

wieset:
Fixed now, thanks for reviewing!

CMakeLists.txt
```diff
@@ -34,6 +34,10 @@ if(WIN32)
   target_compile_definitions(${PROJECT_NAME}
     PRIVATE "RCPPUTILS_BUILDING_LIBRARY")
 endif()
+set_target_properties(${PROJECT_NAME} PROPERTIES
+  BUILD_RPATH ${CMAKE_CURRENT_BINARY_DIR}
+  INSTALL_RPATH_USE_LINK_PATH TRUE
+)
```
clalancette, line 40: The overall thing that concerns me about this is this bit.  As far as I know, we don't currently set RPATH/RUNPATH on any of our shared objects.  In order for this to truly work in general, we would have to start setting that, correct?

wieset: Sorry for taking so long to reply. I think you are right, for this to work in general, you would need to have `set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)` for all projects.

What we are currently doing in our project is using `set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)` together with `set(CMAKE_EXE_LINKER_FLAGS "-Wl,--disable-new-dtags")`. This makes use of RPATH instead of RUNPATH. RPATH works with indirect shared library dependencies, whereas RUNPATH does not. This, together with the changes from this pull request, allows us to load all shared libraries without having to rely on `LD_LIBRARY_PATH`. However, RPATH is deprecated, so I don't think this is a good general solution.

Would it be an option to `set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)` globally for ament_cmake? I am not sure about any downsides of this.

wieset: There has been no activity here in quite a while, so I just wanted to ask if there is any update on this. As a workaround, we are currently running our node in a root shell, but of course we would prefer not to.

spiderkeys: Also wanted to check on the status of this. I'm working on a linuxdeploy plugin for building AppImages that bundle a ROS2 app and its dependencies (similar to the linuxdeploy qt plugin). 

Certain dependencies, like RMW implementations, are searched for at runtime with RPATH/RUNPATH excluded, so even if the executable's RPATH points to the place where these libraries are located, find_library_path() fails to pick them up. Would be nice to not have to rely on setting environment variables like LD_LIBRARY_PATH or having to source the setup.bash scripts as a custom init step in the AppImage's entrypoint (they are already slow enough to start, due to the squashfs mounting step :slightly_smiling_face:)

clalancette: Overall, I think there are two different things going on in this PR (please correct me if I'm wrong):

1.  We are adding code to search the RPATH/RUNPATH for symbols (these are the changes in `find_library.cpp`).
1.  We are switching the code to force RPATH instead of RUNPATH (this is the initial changes in CMakeLists.txt).

I'm fine with the first one; in our current system, that will really have no effect, but gets us closer to what @wieset and others want.  I'm much less sure about the implications about the second one, which I think is what is mostly holding this up.

So my suggestion here is that we split out the second part from this PR, and just concentrate on getting the "search in RPATH" changes in.  Once that is in, we can consider the "force RUNPATH -> RPATH" separately.  Does that make sense?

(also, this PR needs a rebase since things have changed in the meantime)

wieset: True, there are two things going on. I think the first one is obsolete now with https://github.com/ros2/rcutils/pull/320 which is great! The second one is actually a bit different. It does not force `RPATH` instead of `RUNPATH` at all. Even though CMake uses `RPATH` in their variable names, it refers to `RUNPATH` since `RPATH` is deprecated.

- `BUILD_RPATH ${CMAKE_CURRENT_BINARY_DIR}` is necessary so that `RUNPATH` in the build binary gets populated with the path to the test library
- `INSTALL_RPATH_USE_LINK_PATH TRUE` is necessary so that `RUNPATH` in the installed binary gets populated properly

However, thinking about it again, `INSTALL_RPATH_USE_LINK_PATH TRUE` needs to be set for ALL packages so that `RUNPATH` can be used as an alternative to `LD_LIBRARY_PATH`. The reason for this is that `RUNPATH` only works for direct shared object dependencies. See also the discussion here: https://bugs.launchpad.net/ubuntu/+source/glibc/+bug/1737608

And this is what's necessary for us to be able to run a node with `setcap` since it drops `LD_LIBRARY_PATH`. A solution I see would be to have `set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)` in the `ament_package()` macro.

I hope this explanation is clear. What do you think about closing this PR and continuing the discussion in https://github.com/ament/ament_cmake/?

clalancette: > I hope this explanation is clear. What do you think about closing this PR and continuing the discussion in https://github.com/ament/ament_cmake/?

That sounds reasonable to me, thanks.  I'll go ahead and close this; please feel free to open an issue over there.

PR #46, opened 2020-02-29
Add more documentation and include doxyfile

brawner:
To help bring this package up to quality level 1, this PR includes additional doxygen comment blocks for public functionality. Also included is a doxyfile.

Signed-off-by: Stephen Brawner <brawner@gmail.com>

### DIFF
--- a/include/rcpputils/filesystem_helper.hpp
+++ b/include/rcpputils/filesystem_helper.hpp
@@ -33,9 +33,10 @@
 // This file is originally from:
 // https://github.com/ros/pluginlib/blob/1a4de29fa55173e9b897ca8ff57ebc88c047e0b3/pluginlib/include/pluginlib/impl/filesystem_helper.hpp
 
-/**
- * If std::filesystem is not available the necessary functions are emulated.
+/*! \file filesystem_helper.hpp
+ * \brief Cross-platform filesystem helper functions and additional emulation of [std::filesystem](https://en.cppreference.com/w/cpp/filesystem). 
  *
+ * If std::filesystem is not available the necessary functions are emulated.
  * Note: Once std::filesystem is supported on all ROS2 platforms, this class
  * can be deprecated in favor of the built-in functionality.
  */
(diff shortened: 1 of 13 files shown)


include/rcpputils/filesystem_helper.hpp
```diff
@@ -168,26 +202,51 @@ class path
     }
   }
 
+  /**
+  * \brief Check if the path is empty.
```
thomas-moulard:
```suggestion
  * \brief Checks if the path is empty.
```

emersonknapp: I don't think that suggestion is correct, the suggested brief syntax is imperative mood, is it not?

wjwwood: GSG uses the "indicative mood" which makes @thomas-moulard's suggestion in line with that.

However, we commonly use the "imperative mood", probably due to influence from Python's PEP257 (which suggests to use that).

I believe it was discussed on a recent issue:

https://github.com/ros2/rmw/pull/186#discussion_r363918807

Looks like there was split preference for following GSG versus using imperative mood everywhere (even across languages) for consistency.

There seems to be consensus that within a package it should be consistent, though it's not clear what you do in a package which has python and C++ code, because consistency would be at odds with following the idiomatic style for each language.

For this case, I'd recommend doing what's already being done in `rcpputils` elsewhere.

If you guys want to raise this issue more, I'd recommend a pr to the developer guide, clarifying this one way or the other. It's definitely something we're not doing well at the moment ecosystem wide.

emersonknapp: Thanks for that extra context. I don't like strongly enforcing either way if we don't have a linter for it. For now I'd agree and say that we should be consistent within this package.

wjwwood: I tend to agree. That might be partly why we default to imperative, since in Python it is enforced with the pep257 linter. I don't think there is a similar linter for C++ doxygen comments. It's not part of `cpplint.py` for example.

brawner: Cleaned them up and made them consistent. It looks like this project was already used both styles to begin with. It probably doesn't help that the code is combined from several different sources. So I'm choosing to use imperative as that was beaten into me by pep257's linter when porting rqt.
```

brawner:
This PR passes tests and has two approvals if someone would like to merge it (unless of course someone else has comments).

PR #49, opened 2020-03-20
Included abstraction for rcutils::shared_library

ahcorde:
Included abstraction for rcutils::shared_library. To be able to use it with C++ 

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,7 +19,7 @@ if(NOT WIN32)
   add_compile_options(-Wall -Wextra -Wpedantic)
 endif()
 
-option(DISABLE_SANITIZERS "disables the use of gcc saniztizers")
+option(DISABLE_SANITIZERS "disables the use of gcc sanitizers")
 # Only enable sanitizers on x64 architectures
 # https://github.com/google/sanitizers/issues/794
 if(NOT ${CMAKE_SIZEOF_VOID_P} EQUAL 8)
@@ -32,8 +32,8 @@ ament_export_include_directories(include)
 
 add_library(${PROJECT_NAME}
   src/asserts.cpp
-  src/shared_library.cpp
-  src/find_library.cpp)
+  src/find_library.cpp
+  src/shared_library.cpp)
 target_include_directories(${PROJECT_NAME}
   PUBLIC
   include
@@ -78,11 +78,7 @@ if(BUILD_TESTING)
 
   ament_add_gtest(test_pointer_traits test/test_pointer_traits.cpp)
 
-  if(WIN32)
-    set(append_library_dirs "$<TARGET_FILE_DIR:${PROJECT_NAME}>")
-  else()
-    set(append_library_dirs "${CMAKE_CURRENT_BINARY_DIR}")
-  endif()
+  set(append_library_dirs "$<TARGET_FILE_DIR:${PROJECT_NAME}>")
 
   ament_add_gtest(test_shared_library test/test_shared_library.cpp
     APPEND_LIBRARY_DIRS "${append_library_dirs}"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,6 +25,7 @@ ament_export_include_directories(include)
 
 add_library(${PROJECT_NAME}
   src/asserts.cpp
+  src/shared_library.cpp
   src/find_library.cpp)
 target_include_directories(${PROJECT_NAME}
   PUBLIC
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,6 +79,11 @@ if(BUILD_TESTING)
 
   if(TARGET test_shared_library)
     add_library(dummy_shared_library test/dummy_shared_library/dummy_shared_library.c)
+    if(WIN32)
+      # Causes the visibility macros to use dllexport rather than dllimport
+      # which is appropriate when building the dll but not consuming it.
+      target_compile_definitions(dummy_shared_library PRIVATE "DUMMY_SHARED_LIBRARY_BUILDING_DLL")
+    endif()
     target_link_libraries(test_shared_library ${PROJECT_NAME})
   endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,6 +19,13 @@ if(NOT WIN32)
   add_compile_options(-Wall -Wextra -Wpedantic)
 endif()
 
+option(DISABLE_SANITIZERS "disables the use of gcc saniztizers")
+# Only enable sanitizers on x64 architectures
+# https://github.com/google/sanitizers/issues/794
+if(NOT ${CMAKE_SIZEOF_VOID_P} EQUAL 8)
+  set(DISABLE_SANITIZERS ON)
+endif()
+
 include_directories(include)
 
 ament_export_include_directories(include)
@@ -69,6 +76,25 @@ if(BUILD_TESTING)
   ament_add_gtest(test_find_and_replace test/test_find_and_replace.cpp)
 
   ament_add_gtest(test_pointer_traits test/test_pointer_traits.cpp)
+
+  if(WIN32)
+    set(append_library_dirs "$<TARGET_FILE_DIR:${PROJECT_NAME}>")
+  else()
+    set(append_library_dirs "${CMAKE_CURRENT_BINARY_DIR}")
+  endif()
+
+  ament_add_gtest(test_shared_library test/test_shared_library.cpp
+    APPEND_LIBRARY_DIRS "${append_library_dirs}"
+  )
+
+  if(TARGET test_shared_library)
+    add_library(dummy_shared_library test/dummy_shared_library/dummy_shared_library.c)
+    target_link_libraries(test_shared_library ${PROJECT_NAME})
+    if(NOT DISABLE_SANITIZERS)
+      target_compile_options(test_shared_library PUBLIC $<$<CXX_COMPILER_ID:GNU>:-fsanitize=leak>)
+      target_link_libraries(test_shared_library $<$<CXX_COMPILER_ID:GNU>:-fsanitize=leak>)
+    endif()
+  endif()
 
   ament_add_gtest(test_endian test/test_endian.cpp)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,7 +19,7 @@ if(NOT WIN32)
   add_compile_options(-Wall -Wextra -Wpedantic)
 endif()
 
-option(DISABLE_SANITIZERS "disables the use of gcc sanitizers")
+option(DISABLE_SANITIZERS "disables the use of gcc sanitizers" OFF)
 # Only enable sanitizers on x64 architectures
 # https://github.com/google/sanitizers/issues/794
 if(NOT ${CMAKE_SIZEOF_VOID_P} EQUAL 8)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,13 +19,6 @@ if(NOT WIN32)
   add_compile_options(-Wall -Wextra -Wpedantic)
 endif()
 
-option(DISABLE_SANITIZERS "disables the use of gcc sanitizers" OFF)
-# Only enable sanitizers on x64 architectures
-# https://github.com/google/sanitizers/issues/794
-if(NOT ${CMAKE_SIZEOF_VOID_P} EQUAL 8)
-  set(DISABLE_SANITIZERS ON)
-endif()
-
 include_directories(include)
 
 ament_export_include_directories(include)
@@ -87,10 +80,6 @@ if(BUILD_TESTING)
   if(TARGET test_shared_library)
     add_library(dummy_shared_library test/dummy_shared_library/dummy_shared_library.c)
     target_link_libraries(test_shared_library ${PROJECT_NAME})
-    if(NOT DISABLE_SANITIZERS)
-      target_compile_options(test_shared_library PUBLIC $<$<CXX_COMPILER_ID:GNU>:-fsanitize=leak>)
-      target_link_libraries(test_shared_library $<$<CXX_COMPILER_ID:GNU>:-fsanitize=leak>)
-    endif()
   endif()
 
   ament_add_gtest(test_endian test/test_endian.cpp)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,6 +44,7 @@ if(WIN32)
 endif()
 ament_target_dependencies(${PROJECT_NAME} rcutils)
 ament_export_libraries(${PROJECT_NAME})
+ament_export_dependencies(rcutils)
 
 if(BUILD_TESTING)
   find_package(ament_cmake_gtest REQUIRED)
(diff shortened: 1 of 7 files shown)


include/rcpputils/shared_library.hpp
```diff
@@ -0,0 +1,62 @@
...
+#include <stdexcept>
+
+#include "rcutils/shared_library.h"
+
+namespace rcpputils
+{
+
+/**
+ * This class is an abstraction of rcutils shared library to be able to used it
+ *  with modern C++
```
wjwwood: nitpick: punctuation

include/rcpputils/shared_library.hpp
```diff
@@ -0,0 +1,62 @@
...
+ **/
+class SharedLibrary
+{
+public:
+  /// The library is loaded in the constructor
+  /**
+   * \param[in] input The library string path.
+   */
+  explicit SharedLibrary(std::string library_path);
```
wjwwood: Why not `const std::string &`?

include/rcpputils/shared_library.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+  /// The library is unloaded in the deconstructor
+  ~SharedLibrary();
```
wjwwood: Should this be `virtual` or no?

include/rcpputils/shared_library.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+  /// Return true if the shared library contains a specific symbol name otherwise returns false.
+  /**
+   * \param[in] symbol_name name of the symbol inside the shared library
+   * \return if symbols exists returns true, otherwise returns false.
+   */
+  bool has_symbol(std::string symbol_name);
```
wjwwood: Why not const std::string &?

include/rcpputils/shared_library.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+  /// Return shared library symbol pointer.
+  /**
+   * \param[in] symbol_name name of the symbol inside the shared library
+   * \return shared library symbol pointer, if the symbol doesn't exist then returns NULL.
+   */
+  void * get_symbol(std::string symbol_name);
```
wjwwood: Why not const std::string &?

src/shared_library.cpp
```diff
@@ -0,0 +1,61 @@
...
+  }
+}
+
+void * SharedLibrary::get_symbol(std::string symbol_name)
+{
+  void * lib_symbol = rcutils_get_symbol(&lib, symbol_name.c_str());
+
+  if (!lib_symbol) {
+    throw std::runtime_error{"rcpputils SharedLiibrary exception: symbol_name could not be found:" +
+            symbol_name + "in " + lib.library_path};
```
wjwwood:
```suggestion
            symbol_name + " in " + lib.library_path};
```

CMakeLists.txt
```diff
@@ -19,12 +19,20 @@ if(NOT WIN32)
   add_compile_options(-Wall -Wextra -Wpedantic)
 endif()
 
+option(DISABLE_SANITIZERS "disables the use of gcc saniztizers")
```
dirk-thomas: Please specify the default value explicitly.

Spelling: saniztizers

CMakeLists.txt
```diff
@@ -19,12 +19,20 @@ if(NOT WIN32)
...
+  set(DISABLE_SANITIZERS ON)
+endif()
+
 include_directories(include)
 
 ament_export_include_directories(include)
 
 add_library(${PROJECT_NAME}
   src/asserts.cpp
+  src/shared_library.cpp
```
dirk-thomas: Nitpick: insert in alphabetical order.

CMakeLists.txt
```diff
@@ -69,6 +78,25 @@ if(BUILD_TESTING)
 
   ament_add_gtest(test_pointer_traits test/test_pointer_traits.cpp)
 
+  if(WIN32)
+    set(append_library_dirs "$<TARGET_FILE_DIR:${PROJECT_NAME}>")
+  else()
+    set(append_library_dirs "${CMAKE_CURRENT_BINARY_DIR}")
```
dirk-thomas: Same as https://github.com/ros2/rcutils/pull/215#discussion_r396550326

include/rcpputils/shared_library.hpp
```diff
@@ -0,0 +1,69 @@
...
+ * This class is an abstraction of rcutils shared library to be able to used it
+ *  with modern C++.
+ **/
+class RCPPUTILS_PUBLIC SharedLibrary
+{
+public:
+  /// The library is loaded in the constructor.
+  /**
+   * \param[in] input The library string path.
+   */
```
dirk-thomas: The docblock should mentioned the potentially raised exceptions.

include/rcpputils/shared_library.hpp
```diff
@@ -0,0 +1,69 @@
...
+   * \return if symbols exists returns true, otherwise returns false.
+   */
+  RCPPUTILS_PUBLIC
+  bool
+  has_symbol(const std::string & symbol_name);
+
+  /// Return shared library symbol pointer.
+  /**
+   * \param[in] symbol_name name of the symbol inside the shared library
+   * \return shared library symbol pointer, if the symbol doesn't exist then returns NULL.
```
dirk-thomas: The doc doesn't match the implementation. It currently raises an exception instead of returning `NULL`.

test/test_shared_library.cpp
```diff
@@ -0,0 +1,36 @@
...
+  EXPECT_ANY_THROW(std::make_shared<rcpputils::SharedLibrary>("library_path"));
+
+  auto library = std::make_shared<rcpputils::SharedLibrary>(library_path);
+
+  EXPECT_TRUE(library->has_symbol("print_name"));
+  EXPECT_FALSE(library->has_symbol("symbol"));
+
+  EXPECT_TRUE(library->get_symbol("print_name") != NULL);
+  EXPECT_ANY_THROW(library->get_symbol("symbol"));
+}
```
dirk-thomas: The tests should also exercise non-standard sequences of function calls, e.g. check for expected exceptions.

src/shared_library.cpp
```diff
@@ -0,0 +1,64 @@
...
+  lib = rcutils_get_zero_initialized_shared_library();
+  rcutils_ret_t ret = rcutils_load_shared_library(
+    &lib,
+    library_path.c_str(),
+    rcutils_get_default_allocator());
+  if (ret != RCUTILS_RET_OK) {
+    if (ret == RCUTILS_RET_BAD_ALLOC) {
+      throw std::runtime_error{"rcutils shared_library exception: failed to allocate memory"};
+    } else if (ret == RCUTILS_RET_INVALID_ARGUMENT) {
+      throw std::runtime_error{"rcutils shared_library exception: invalid arguments"};
```
dirk-thomas: How can this be the case?

wjwwood: Doesn't look like it can be, but it should still be checked no? What if there's a bug in the underlying implementation that causes this error?

dirk-thomas: Yeah, checking it nevetheless is good. Maybe just add a comment that this shouldn't be possible.

CMakeLists.txt
ahcorde: The specified value is two lines below

dirk-thomas: No, two lines below the value is conditionally changed to `ON`. I want the default value to be specified in this line as the third argument as `OFF`.

test/test_shared_library.cpp
ahcorde: https://github.com/ros2/rcpputils/pull/49/commits/4a128d9513d505d0f85e61bd9626fe154cdc3247

CMakeLists.txt
dirk-thomas: See 4a128d9.

test/test_shared_library.cpp
ahcorde: There is no option to tests get_libray_path, get_symbol or has_symbol when the library doesn't exist, because the constructor throws exceptions and there is an explicit constructor

https://github.com/ros2/rcpputils/pull/49/commits/5630628abadf79b87386607ae6676999df22b405

ahcorde:
friendly ping @dirk-thomas @wjwwood

src/shared_library.cpp
```diff
@@ -0,0 +1,72 @@
...
+              library_path};
+    }
+  }
+}
+
+SharedLibrary::~SharedLibrary()
+{
+  rcutils_ret_t ret = rcutils_unload_shared_library(&lib);
+  if (ret != RCUTILS_RET_OK) {
+    std::cerr << "rcutils shared_library exception: failed destroying object" << std::endl;
```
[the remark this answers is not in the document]
ahcorde: you mean that I should call `rcutils_reset_error()` before throw the exception?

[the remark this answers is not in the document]
wjwwood: Yes. You should also consider using the error message set from the called function rather than repeating then here.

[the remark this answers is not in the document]
ahcorde: Included  `rcutils_reset_error` and `rcutils_get_error_string`: https://github.com/ros2/rcpputils/pull/49/commits/efc99631129947595bcdeee4923b38f5f5f9903b

CMakeLists.txt
```diff
@@ -19,13 +19,21 @@ if(NOT WIN32)
   add_compile_options(-Wall -Wextra -Wpedantic)
 endif()
 
+option(DISABLE_SANITIZERS "disables the use of gcc sanitizers" OFF)
+# Only enable sanitizers on x64 architectures
+# https://github.com/google/sanitizers/issues/794
+if(NOT ${CMAKE_SIZEOF_VOID_P} EQUAL 8)
+  set(DISABLE_SANITIZERS ON)
+endif()
+
```
dirk-thomas: Imo this new option is completely orthogonal to the feature implemented in this PR. Therefore it should be moved into a separate PR. Maybe it would make sense to add this option to `ament_cmake_ros` instead?

include/rcpputils/shared_library.hpp
```diff
@@ -0,0 +1,82 @@
...
+   * \throws std::bad_alloc if allocating storage for the callback fails
+   * \throws std::runtime_error if there are some invalid arguments or the library
+   * was not load properly
+   */
+  RCPPUTILS_PUBLIC
+  explicit SharedLibrary(const std::string & library_path);
+
+  /// The library is unloaded in the deconstructor
+  RCPPUTILS_PUBLIC
+  virtual ~SharedLibrary();
```
dirk-thomas: Why is (only) the destructor virtual?

src/shared_library.cpp
```diff
@@ -0,0 +1,75 @@
...
+{
+  return rcutils_has_symbol(&lib, symbol_name.c_str());
+}
+
+std::string SharedLibrary::get_library_path()
+{
+  if (lib.library_path != nullptr) {
+    return std::string(lib.library_path);
+  }
+  return std::string("");
```
dirk-thomas: Would it make more sense to raise an exception? Also this behavior is not documented i the docblock of the method.

include/rcpputils/shared_library.hpp
emersonknapp: I think they're just trying to follow C++ best practices, which is "all destructors should be virtual". Even if this is not intended to be subclassed, somebody could do so, and it should clean itself up, which will not happen if the dtor is not virtual.

wjwwood: Indeed, if you wanted to make it non virtual then you could mark the class final or make the destructor protected so no one could try to delete a base class pointer to this.

src/shared_library.cpp
ahcorde: the method throws an exception, the docblock was also update https://github.com/ros2/rcpputils/pull/49/commits/da90788662b73d53b996a21715f6e5bea698e815

CMakeLists.txt
ahcorde: Removed sanitizers https://github.com/ros2/rcpputils/pull/49/commits/2d5338096c60621b95e6e71d38cbec5e5fd1d5d6

test/test_shared_library.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+#include <gtest/gtest.h>
+
+#include <memory>
+#include <string>
+
+#include "rcpputils/shared_library.hpp"
+
+TEST(test_shared_library, valid_load) {
+  // library path
```
dirk-thomas: Nitpick: I would remove this comment since it adds no value.

test/test_shared_library.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+TEST(test_shared_library, failed_test) {
+  // Trying to reset the object two times
+  try {
+    std::shared_ptr<rcpputils::SharedLibrary> library;
+    library.reset();
+    library.reset();
+  } catch (...) {
+    FAIL();
+  }
```
dirk-thomas: This isn't really testing anything of `SharedLibrary` but the `shared_ptr` Therefore I would suggest to remove it (or at least remove the second `reset()` call and the comment if we want to just use the construction / destruction which is also done below).

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9871/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5487/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8031/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9733/)

CI is green @wjwwood or @dirk-thomas can you review last changes?

include/rcpputils/shared_library.hpp
```diff
@@ -0,0 +1,93 @@
...
+   */
+  RCPPUTILS_PUBLIC
+  bool
+  has_symbol(const std::string & symbol_name);
+
+  /// Return shared library symbol pointer.
+  /**
+   * \param[in] symbol_name name of the symbol inside the shared library
+   * \return shared library symbol pointer, if the symbol doesn't exist then throws a
+   * \runtime_error exception
```
wjwwood, line 61: nitpick: what does the `\runtime_error` do here? Is that a doxygen command or escape sequence or something?

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9877/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5495/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8036/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9735/)

include/rcpputils/shared_library.hpp
```diff
@@ -58,7 +58,7 @@ class SharedLibrary
   /**
    * \param[in] symbol_name name of the symbol inside the shared library
    * \return shared library symbol pointer, if the symbol doesn't exist then throws a
-   * \runtime_error exception
+   * runtime_error exception
```
emersonknapp, line 61: It seems like "if the symbol doesn't exist then throws a runtime_error exception" isn't necessary to say, since you repeat it on the next line.

PR #50, opened 2020-04-03
Added unload_library method to shared_library

ahcorde:
As discussed in the PR https://github.com/ros/class_loader/pull/139#pullrequestreview-387397419 I will include a unload method to the sharedLibrary abstraction

### DIFF
--- a/src/shared_library.cpp
+++ b/src/shared_library.cpp
@@ -41,20 +41,11 @@ SharedLibrary::SharedLibrary(const std::string & library_path)
 
 SharedLibrary::~SharedLibrary()
 {
-  if (lib.lib_pointer != NULL) {
+  if (rcutils_is_shared_library_loaded(&lib)) {
     rcutils_ret_t ret = rcutils_unload_shared_library(&lib);
     if (ret != RCUTILS_RET_OK) {
       std::cerr << rcutils_get_error_string().str << std::endl;
       rcutils_reset_error();
-    }
-  } else {
-    if (lib.library_path != NULL) {
-      if (rcutils_allocator_is_valid(&lib.allocator)) {
-        lib.allocator.deallocate(lib.library_path, lib.allocator.state);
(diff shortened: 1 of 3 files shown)


dirk-thomas:
Have you tried this patch? I would assume if you invoke `unload_library` first that the constructor will fail since it will try to unload the library again.

src/shared_library.cpp
```diff
@@ -40,11 +40,32 @@ SharedLibrary::SharedLibrary(const std::string & library_path)
...
+  } else {
+    if (lib.library_path != NULL) {
+      if (rcutils_allocator_is_valid(&lib.allocator)) {
+        lib.allocator.deallocate(lib.library_path, lib.allocator.state);
+        lib.allocator = rcutils_get_zero_initialized_allocator();
+      } else {
+        std::cerr << "ShareLibrary not able to deallocate memory on constructor" << std::endl;
+      }
+    }
+  }
```
dirk-thomas, line 59: Why does this logic need to be here? Can't this call the unload API when a certain condition is met instead? Having knowledge about how the stuff needs to be deallocated etc. seems to break encapsulation.

ahcorde: fixed https://github.com/ros2/rcpputils/pull/50/commits/cb11c673b72e18723647486e19bce1c06dcace72

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9983/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5583/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8122/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9858/)

ahcorde:
Openning and clossing the issue to retest the github actions

ahcorde:
Openning and clossing the issue to retest the github actions

PR #51, opened 2020-04-08
Increased shared library tests

ahcorde:
Increased shared library tests

### DIFF
--- a/include/rcpputils/shared_library.hpp
+++ b/include/rcpputils/shared_library.hpp
@@ -80,13 +80,15 @@ private:
 
 /// Get the platform specific library name
 /**
+ * The maximun file name size is 1024 characters, if the input library_name is bigger than
+ * this value then the method returns an empty string
+ *
  * \param[in] symbol_name library base name (without prefix and extension)
  * \return platform specific library name
  * \throws std::runtime_error if it's not able to create the library name
  */
 RCPPUTILS_PUBLIC
 std::string get_platform_library_name(std::string library_name);
-
 
 }  // namespace rcpputils
 
--- a/include/rcpputils/shared_library.hpp
+++ b/include/rcpputils/shared_library.hpp
@@ -89,7 +89,7 @@ private:
 /// Get the platform specific library name
 /**
  * The maximun file name size is 1024 characters, if the input library_name is bigger than
- * this value then the method returns an empty string
+ * this value then the method throws an exception.
  *
  * \param[in] symbol_name library base name (without prefix and extension)
  * \param[in] debug if true the library will return a debug library name, otherwise
--- a/include/rcpputils/shared_library.hpp
+++ b/include/rcpputils/shared_library.hpp
@@ -88,7 +88,7 @@ private:
 
 /// Get the platform specific library name
 /**
- * The maximun file name size is 1024 characters, if the input library_name is bigger than
+ * The maximum file name size is 1024 characters, if the input library_name is bigger than
  * this value then the method throws an exception.
  *
  * \param[in] symbol_name library base name (without prefix and extension)
--- a/src/shared_library.cpp
+++ b/src/shared_library.cpp
@@ -76,6 +76,7 @@ std::string SharedLibrary::get_library_path()
 std::string get_platform_library_name(std::string library_name)
 {
   char library_name_platform[1024];
+  memset(library_name_platform, 0, 1024);
   rcutils_ret_t ret = rcutils_get_platform_library_name(
     library_name.c_str(),
     library_name_platform,
--- a/src/shared_library.cpp
+++ b/src/shared_library.cpp
@@ -75,8 +75,7 @@ std::string SharedLibrary::get_library_path()
 
 std::string get_platform_library_name(std::string library_name)
 {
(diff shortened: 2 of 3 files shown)


ahcorde:
Right now it's bigger than 95%

src/shared_library.cpp
```diff
@@ -76,6 +76,7 @@ std::string SharedLibrary::get_library_path()
 std::string get_platform_library_name(std::string library_name)
 {
   char library_name_platform[1024];
+  memset(library_name_platform, 0, 1024);
```
wjwwood, line 79: Using `std::array` here would be more idiomatic C++:

```c++
std::array<char, 1024> library_name_platform{};
```

Either way, you could use "value initialization" to make sure it is zeroed without using `memset`:

```c++
char library_name_platform[1024]{};
```

See: https://stackoverflow.com/a/18295840/671658

If we had C++17, then you could have used `std::string library_name_platform(1024, '\0');`, but the non-const `std::basic_string::data` method wasn't added until C++17...

test/test_shared_library.cpp
```diff
@@ -55,3 +55,14 @@ TEST(test_shared_library, failed_test) {
     FAIL();
   }
 }
+
+TEST(test_get_platform_library_name, failed_test) {
+  // create a string bigger than the internal buffer
+  int length = 2000;
+  std::string str;
```
wjwwood, line 62: Instead:

```c++
std::string str(2000, 'A');
```

test/test_shared_library.cpp
```diff
@@ -55,3 +55,14 @@ TEST(test_shared_library, failed_test) {
...
+  for (int i = 0; i < length; i++) {
+    str.push_back('A');
+  }
+  std::string library_path = rcpputils::get_platform_library_name(str);
+  EXPECT_TRUE(library_path.empty());
```
wjwwood, line 67: This behavior (returning an empty string) doesn't seem to be documented, or am I missing the point of this test? Is it because the input str doesn't exist locally (it seems like that shouldn't matter)? I don't understand, even after looking at the rcutils implementation.

ahcorde: The idea it's to use a library_name bigger than 1024 characters ( the size of the internal buffer ), then `rcutils_get_platform_library_name` should return a `RCUTILS_RET_ERROR`

I will document about the maximum size and empty string

wjwwood: But if there's an rcutils error code, will this not throw... I don't see any case where the string is empty.

ahcorde: That's true :man_facepalming:

wjwwood: This may be related to how you're using `rcutils_snprintf`? The input being too long is not an error case, it is supposed to truncate the string instead, see:

https://github.com/ros2/rcutils/blob/04456c7540362cf60a01b19d0e505fded4156795/test/test_snprintf.cpp#L32-L34

ahcorde: As far as I see, there is not tests in which the buffer is smaller than the string to be copied. which is the case that I'm trying to cover

include/rcpputils/shared_library.hpp
```diff
@@ -88,8 +88,11 @@ class SharedLibrary
 
 /// Get the platform specific library name
 /**
-* \param[in] symbol_name library base name (without prefix and extension)
-* \param[in] debug if true the library will return a debug library name, otherwise
+ * The maximun file name size is 1024 characters, if the input library_name is bigger than
```
wjwwood, line 91:
```suggestion
 * The maximum file name size is 1024 characters, if the input library_name is bigger than
```

include/rcpputils/shared_library.hpp
```diff
@@ -88,8 +88,11 @@ class SharedLibrary
...
+ * this value then the method returns an empty string
```
wjwwood, line 92:
```suggestion
 * this value then the method returns an empty string.
```

test/test_shared_library.cpp
```diff
@@ -64,3 +64,9 @@ TEST(test_shared_library, failed_test) {
     FAIL();
   }
 }
+
+TEST(test_get_platform_library_name, failed_test) {
+  // create a string bigger than the internal buffer
+  std::string str(2000, 'A');
+  EXPECT_THROW(rcpputils::get_platform_library_name(str), std::runtime_error);
```
wjwwood, line 71: This now checks for a throw, but the documentation still says it returns an empty string...

include/rcpputils/shared_library.hpp
wjwwood: This is still pending.

ahcorde:
Running the CI up-to rcpputils and testing only rcpputils because this is a PR that only affects to tests

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10221/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5769/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8309/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10109/)

PR #63, opened 2020-05-18
Add current_path to filesystem_helpers

cottsay:
Modeled after the C++ 17 function: https://en.cppreference.com/w/cpp/filesystem/current_path

I decided not to use the `rcutils_get_cwd` so that error information could be reliably propagated and so that the buffer could be dynamically allocated.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11020/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6352/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8977/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10947/)

### DIFF
--- a/include/rcpputils/filesystem_helper.hpp
+++ b/include/rcpputils/filesystem_helper.hpp
@@ -44,6 +44,7 @@
 #ifndef RCPPUTILS__FILESYSTEM_HELPER_HPP_
 #define RCPPUTILS__FILESYSTEM_HELPER_HPP_
 
+#include <limits.h>
 #include <sys/stat.h>
 
 #include <algorithm>
@@ -421,24 +422,22 @@ inline path temp_directory_path()
  */
 inline path current_path()
 {
-  char * cwd;
 #ifdef _WIN32
 #ifdef UNICODE
 #error "rcpputils::fs does not support Unicode paths"
 #endif
-  cwd = _getcwd(NULL, 0);
+  char cwd[MAX_PATH];
+  if (nullptr == _getcwd(cwd, MAX_PATH)) {
 #else
-  cwd = getcwd(NULL, 0);
+  char cwd[PATH_MAX];
+  if (nullptr == getcwd(cwd, PATH_MAX)) {
 #endif
-  if (nullptr == cwd) {
     std::error_code ec{errno, std::system_category()};
     errno = 0;
     throw std::system_error{ec, "cannot get current working directory"};
   }
 
-  std::string ret(cwd);
-  free(cwd);
-  return path(ret);
+  return path(cwd);
 }
 
 /**
--- a/include/rcpputils/filesystem_helper.hpp
+++ b/include/rcpputils/filesystem_helper.hpp
@@ -413,6 +413,35 @@ inline path temp_directory_path()
 }
 
 /**
+ * \brief Return current working directory.
+ *
+ * \return The current working directory.
+ *
+ * \throws std::system_error
+ */
+inline path current_path()
+{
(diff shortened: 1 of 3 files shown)


dirk-thomas:
Please briefly mention what this new API is needed / used for.

test/test_filesystem_helper.cpp
```diff
@@ -229,3 +229,9 @@ TEST(TestFilesystemHelper, remove_extension_no_extension)
   p = rcpputils::fs::remove_extension(p);
   EXPECT_EQ("foo", p.string());
 }
+
+TEST(TestFilesystemHelper, get_cwd)
```
ahcorde, line 233: can we add here (for example) an env var to check that the current directory returned is the right value?

cottsay:
> Please briefly mention what this new API is needed / used for.

Take a look at the tests I'm proposing for `rcl_logging_spdlog`: https://github.com/ros2/rcl_logging/pull/40/files

I'm creating a dummy directory in the test's working directory (which should be somewhere in the `build/rcl_logging_spdlog/` directory) to act as a temporary "home directory". The code I'm testing doesn't function correctly if that directory path isn't an absolute path, so I need to get the current working directory to make that happen.

test/test_filesystem_helper.cpp
cottsay: ed673ad

include/rcpputils/filesystem_helper.hpp
```diff
@@ -412,6 +412,35 @@ inline path temp_directory_path()
...
+#ifdef UNICODE
+#error "rcpputils::fs does not support Unicode paths"
+#endif
+  cwd = _getcwd(NULL, 0);
+#else
+  cwd = getcwd(NULL, 0);
+#endif
+  if (nullptr == cwd) {
+    std::error_code ec{errno, std::system_category()};
+    errno = 0;
```
hidmic, line 435: @cottsay I'd rater not change `errno` in the process.

include/rcpputils/filesystem_helper.hpp
```diff
@@ -412,6 +412,35 @@ inline path temp_directory_path()
...
+inline path current_path()
+{
+  char * cwd;
+#ifdef _WIN32
+#ifdef UNICODE
+#error "rcpputils::fs does not support Unicode paths"
+#endif
+  cwd = _getcwd(NULL, 0);
+#else
+  cwd = getcwd(NULL, 0);
```
hidmic, line 431: @cottsay nit: although all our currently supported platforms support this use of `getcwd()` (i.e. passing in a NULL buffer, yielding a dynamically allocated one), it's not part of POSIX. 

Creating a `PATH_MAX` (or `MAX_PATH` in Windows) long array for a buffer would keep it compliant.

include/rcpputils/filesystem_helper.hpp
cottsay: I'm following the pattern that exists in this file already. If you think that pattern should change, you should start that discussion, but I think that consistent behavior is more important for now.
https://github.com/ros2/rcpputils/blob/eee6c97cbed514afc0ed1160f30a0eb176fd8a0b/include/rcpputils/filesystem_helper.hpp#L197-L199

include/rcpputils/filesystem_helper.hpp
cottsay: 283a2f3

PR #66, opened 2020-06-01
Update quality declaration for version stability.

nuclearsandwich:
1.0 for this package was bumped in https://github.com/ros2/rcpputils/commit/547498373d8cc9f9413c2f9a4a81db91b410787e so this updates the quality declaration.

It looks to me like this might make the package level 2 or level 3 but for the rcutils dependency. I'd appreciate a comment on that front from someone whose got more familiarity with what's part of each quality level.


nuclearsandwich:
> Same as rcutils. Coverage it's not ready. It's QL 3.

Ack. I misread "it's not ready" as no change needed instead of bumping to QL3.

PR #68, opened 2020-06-02
Align path combine behavior with C++17

cottsay:
Combining any path with an absolute path yields only the absolute path in the C++17 implementation.

The current behavior here is to smush them together anyway.

Requires #67

### DIFF
--- a/include/rcpputils/filesystem_helper.hpp
+++ b/include/rcpputils/filesystem_helper.hpp
@@ -326,10 +326,15 @@ public:
   */
   path & operator/=(const path & other)
   {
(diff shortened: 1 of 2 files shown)


cottsay:
Only rcpputils:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10967/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6306/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8928/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10871/)

PR #69, opened 2020-06-02
Include stdexcept in get_env.hpp

cottsay:
Needed for this line:
https://github.com/ros2/rcpputils/blob/6e6a1f3c6a92ad6ad0344dcefc840f86678a5c0e/include/rcpputils/get_env.hpp#L56

Otherwise consumers need to include it themselves to avoid:
```
include/rcpputils/get_env.hpp: In function ‘std::string rcpputils::get_env_var(const char*)’:
include/rcpputils/get_env.hpp:56:16: error: ‘runtime_error’ is not a member of ‘std’
   56 |     throw std::runtime_error(err);
      |                ^~~~~~~~~~~~~
```

### DIFF
--- a/include/rcpputils/get_env.hpp
+++ b/include/rcpputils/get_env.hpp
@@ -35,6 +35,7 @@
 
 #include "rcutils/get_env.h"
 
+#include <stdexcept>
 #include <string>
 
 #include "rcpputils/visibility_control.hpp"


wjwwood:
In this case I think the GitHub action checks are sufficient, so no need for ci.ros2.org CI to be run, i.e. I don't expect this to fail on Ubuntu 20.04, macOS, or Windows.

PR #70, opened 2020-06-04
Add get_executable_name() function

cottsay:
This function wraps `rcutils_get_executable_name`. It takes care of the allocator and throws an exception when an error occurs.

The rcutils_get_executable_name implementation doesn't set any error information when it fails, and from reading through the implementation and the documentation for the system functions it calls, an error should be extremely unlikely. A generic failure message is the best I can do.

Inspired by https://github.com/ros2/rcl_logging/pull/40#discussion_r434196444

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11024/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6356/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8981/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10951/)

### DIFF
--- a/test/test_process.cpp
+++ b/test/test_process.cpp
@@ -19,5 +19,5 @@
 #include "rcpputils/process.hpp"
 
 TEST(TestProcess, test_get_executable_name) {
-  EXPECT_STREQ("test_process", rcpputils::get_executable_name().c_str());
+  EXPECT_EQ("test_process", rcpputils::get_executable_name());
 }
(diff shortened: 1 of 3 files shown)


test/test_process.cpp
```diff
@@ -0,0 +1,23 @@
...
+// limitations under the License.
+
+#include <gtest/gtest.h>
+
+#include <string>
+
+#include "rcpputils/process.hpp"
+
+TEST(TestProcess, test_get_executable_name) {
+  EXPECT_STREQ("test_process", rcpputils::get_executable_name().c_str());
```
hidmic, line 22: @cottsay fyi: I'd expect 
```suggestion
  EXPECT_EQ("test_process", rcpputils::get_executable_name());
```
to work the same.

cottsay: You're totally right, nice catch.

PR #73, opened 2020-06-18
Fix parent_path() for empty paths and paths of length one

jacobperron:
Slightly unrelated, but this PR also adds tests for `exists()` to clarify it's expected behavior (36ebd99).

This fixes an issue discovered downstream: https://github.com/ros2/rosbag2/pull/345/files#r442416506

### DIFF
--- a/include/rcpputils/filesystem_helper.hpp
+++ b/include/rcpputils/filesystem_helper.hpp
@@ -251,6 +251,15 @@ public:
   */
   path parent_path() const
   {
+    // Edge case: if path only consists of one part, then return '.' or '/'
+    //            depending if the path is absolute or not
+    if (1u == path_as_vector_.size()) {
+      if (this->is_absolute()) {
+        return path(std::string(1, kPreferredSeparator));
+      }
+      return path(".");
+    }
+
     path parent;
     for (auto it = this->cbegin(); it != --this->cend(); ++it) {
       if (!parent.empty() || it->empty()) {
--- a/include/rcpputils/filesystem_helper.hpp
+++ b/include/rcpputils/filesystem_helper.hpp
@@ -221,7 +221,8 @@ public:
   bool is_absolute() const
   {
     return path_.size() > 0 &&
-           (path_.compare(0, 1, "/") == 0 || path_.compare(1, 2, ":\\") == 0);
+           (path_.compare(0, 1, std::string(1, kPreferredSeparator)) == 0 ||
+           this->is_absolute_with_drive_letter());
   }
 
   /**
@@ -260,9 +261,19 @@ public:
     //            depending if the path is absolute or not
     if (1u == path_as_vector_.size()) {
       if (this->is_absolute()) {
+        // Windows is tricky, since an absolute path may start with 'C:\\' or '\\'
+        if (this->is_absolute_with_drive_letter()) {
+          return path(path_as_vector_[0] + kPreferredSeparator);
+        }
         return path(std::string(1, kPreferredSeparator));
       }
       return path(".");
+    }
+
+    // Edge case: with a path 'C:\\foo' we want to return 'C:\\' not 'C:'
+    // Don't drop the root directory from an absolute path on Windows starting with a letter drive
+    if (2u == path_as_vector_.size() && this->is_absolute_with_drive_letter()) {
+      return path(path_as_vector_[0] + kPreferredSeparator);
     }
 
     path parent;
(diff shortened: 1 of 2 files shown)


test/test_filesystem_helper.cpp
```diff
@@ -187,6 +205,22 @@ TEST(TestFilesystemHelper, is_empty)
...
+TEST(TestFilesystemHelper, exists)
+{
+  {
+    auto p = path("");
+    EXPECT_FALSE(p.exists());
+  }
+  {
+    auto p = path(".");
+    EXPECT_TRUE(p.exists());
+  }
```
Karsten1987, line 217: * what about `/`
* what about Windows specifics such as`C`, `C:`, `C:\`?

include/rcpputils/filesystem_helper.hpp
```diff
@@ -251,6 +251,20 @@ class path
...
+    }
+
+    // Edge case: if path only consists of one part, then return '.' or '/'
+    //            depending if the path is absolute or not
+    if (1u == path_as_vector_.size()) {
+      if (this->is_absolute()) {
+        return path(std::string(1, kPreferredSeparator));
+      }
+      return path(".");
+    }
```
Karsten1987, line 266: That wouldn't hold on Windows, would it? The parent path of `C:\` is undefined I would say.

test/test_filesystem_helper.cpp
```diff
@@ -70,9 +70,27 @@ TEST(TestFilesystemHelper, join_path)
...
+    auto p = path("my") / path("path");
+    EXPECT_EQ(p.parent_path().string(), path("my").string());
+  }
+  {
+    auto p = path("foo");
+    EXPECT_EQ(p.parent_path().string(), ".");
+  }
+  {
+    if (is_win32) {
+      auto p = path("\\foo");
```
jacobperron, line 83: This test is wrong according to the logic in `is_absolute()`. The prefix should have a single character followed by a colon. e.g. `C:\\`

test/test_filesystem_helper.cpp
jacobperron: Good point. Actually, we could have the same issue on *nix systems too: what is the parent path of "/"?

It seems clearly documented here: https://en.cppreference.com/w/cpp/filesystem/path/parent_path

The parent of "/" is "/". So, I think on Windows we should expect the parent of "C:\" is "C:\".

I've added tests to check for this 9a48203 (I think the logic is correct as-is).

include/rcpputils/filesystem_helper.hpp
jacobperron: Good point. 
It seems clearly documented here: https://en.cppreference.com/w/cpp/filesystem/path/parent_path

The parent of "/" is "/". So, I think on Windows we should expect the parent of "C:\\" is "C:\\".

I've added tests to check for this 9a48203 (I think the logic is correct as-is).

test/test_filesystem_helper.cpp
jacobperron: Fixed in 2f99349

test/test_filesystem_helper.cpp
jacobperron: I don't know if we can assume that a drive with the letter "C" exists, but I think `\\` is also valid on Windows.

See 5c4cd01

jacobperron:
Testing rcpputils:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11207/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6480/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9166/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11124/)

jacobperron:
Fixed expected output syntax on Windows: a2edff8

* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11128/)

jacobperron:
Windows paths were trickier than I anticipated.
The changes in cf12223 handles absolute paths for the current drive (e.g. `\foo` or `\`) and for absolute paths with a letter drive specified (e.g. `C:\foo` or `C:\`)

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11214/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6486/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9174/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11132/)

@Karsten1987 @wjwwood Can you take another look? Thanks!

PR #74, opened 2020-06-18
Fast-forward Foxy

jacobperron:
I believe all commits on `master` can be backported to `foxy`. With approval, I will fast-forward merge `master` into `foxy`. Of all the commits, the only code change not explicitly labeled for backport is #70 (though I don't see any issue in backporting it).

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11219/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6490/)
* macOS [Build Status](https://ci.ros2.org/job/ci_osx/9199/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11136/)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -72,7 +72,11 @@ if(BUILD_TESTING)
 
   ament_add_gtest(test_split test/test_split.cpp)
 
-  ament_add_gtest(test_filesystem_helper test/test_filesystem_helper.cpp)
+  ament_add_gtest(test_filesystem_helper test/test_filesystem_helper.cpp
+    ENV
+      EXPECTED_WORKING_DIRECTORY=$<SHELL_PATH:${CMAKE_CURRENT_BINARY_DIR}>
+    WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
+  ament_target_dependencies(test_filesystem_helper rcutils)
 
   ament_add_gtest(test_find_and_replace test/test_find_and_replace.cpp)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -82,6 +82,9 @@ if(BUILD_TESTING)
 
   ament_add_gtest(test_pointer_traits test/test_pointer_traits.cpp)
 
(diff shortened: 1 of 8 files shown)


emersonknapp:
My only concern is that new APIs were added - is this allowed in distro patch protocol?

wjwwood:
As long as it is ABI compatible then yes. We just also have to bump the minor patch number.

ablasdel:
I think any release adding backwards compatible API requires a MINOR version bump. https://semver.org/

Snipped out relevant bit that is relevant to the conversation:
—-
MAJOR version when you make incompatible API changes,
MINOR version when you add functionality in a backwards compatible manner, and
PATCH version when you make backwards compatible bug fixes.
—-

emersonknapp:
@jacobperron so should we bump the minor version as part of this PR?

clalancette:
> @jacobperron so should we bump the minor version as part of this PR?

No, our release tools will then get confused and try to bump it again later.  Also, in the general case where you have multiple packages in the same repository, the release tools won't work at all (that doesn't apply to this case, since there is only a single package, but I'd like not to make a habit of it).

Unfortunately, we don't currently have a good way of tracking the fact that this package needs a minor bump the next time it is released.  I guess we'll just have to merge and release right away for now.

emersonknapp:
Ok, I've approved, merge when ready

jacobperron:
> @jacobperron so should we bump the minor version as part of this PR?

IMO, we should make a release on `master` now and fast-forward the release commit/tag too. That could be done as part of this PR (or in a follow-up).

wjwwood:
I agree with making the release on master and fast forwarding it to the foxy branch, but I wouldn't put it in this pull request, just because there might be other changes too, so generally it doesn't make sense to include a release with a change.

jacobperron:
I can make a release on `master` and fast-foward it to `foxy` after this is merged.

wjwwood:
Oh, well sorry I was thinking this was a normal feature or bug fix, but since this pr is specifically for fast forwarding, either way would make sense.

PR #79, opened 2020-07-01
filesystem helpers: adding remove_all to remove non-empty directories

### LINKED ISSUE #62: Need for std::filesystem::remove_all equivalent
At least on windows, the `remove()` function fails when the to be deleted directory is non empty.
This requires sort of a recursive deletion as used within rosbag2: https://github.com/ros2/rosbag2/blob/c28c89267fa8e22e76c6540ffb700f908eca04db/rosbag2_test_common/include/rosbag2_test_common/temporary_directory_fixture.hpp#L59-L99

Karsten1987:
fixes #62 

This is a transfer from what we're using at rosbag2: https://github.com/ros2/rosbag2/blob/master/rosbag2_test_common/include/rosbag2_test_common/temporary_directory_fixture.hpp#L59-L99

CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11350/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6582/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9307/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11243/)

### DIFF
--- a/include/rcpputils/filesystem_helper.hpp
+++ b/include/rcpputils/filesystem_helper.hpp
@@ -570,9 +570,7 @@ inline bool remove_all(const path & p)
   auto ret = SHFileOperation(&file_options);
   delete[] temp_dir;
 
-  if (0 != ret) {
-    return false;
-  }
+  return 0 == ret && false == file_options.fAnyOperationsAborted;
 #else
   DIR * dir = opendir(p.string().c_str());
   struct dirent * directory_entry;
@@ -592,9 +590,8 @@ inline bool remove_all(const path & p)
   closedir(dir);
   // directory is empty now, call remove
   remove(p);
-#endif
-
   return !rcpputils::fs::exists(p);
+#endif
 }
 
 /**
--- a/include/rcpputils/filesystem_helper.hpp
+++ b/include/rcpputils/filesystem_helper.hpp
@@ -547,28 +547,32 @@ inline bool remove(const path & p)
  */
 inline bool remove_all(const path & p)
 {
-  if (!is_directory(p)) {return false;}
+  if (!is_directory(p)) {return remove(p);}
 
 #ifdef _WIN32
   // We need a string of type PCZZTSTR, which is a double null terminated char ptr
   size_t length = p.string().size();
   TCHAR * temp_dir = new TCHAR[length + 2];
   memcpy(temp_dir, p.string().c_str(), length);
-  temp_dir[length] = 0;
-  temp_dir[length + 1] = 0;    // double null terminated
+  temp_dir[length] = '\0';
+  temp_dir[length + 1] = '\0';  // double null terminated
 
   SHFILEOPSTRUCT file_options;
   file_options.hwnd = nullptr;
-  file_options.wFunc = FO_DELETE;    // delete (recursively)
+  file_options.wFunc = FO_DELETE;  // delete (recursively)
   file_options.pFrom = temp_dir;
   file_options.pTo = nullptr;
-  file_options.fFlags = FOF_NOCONFIRMATION | FOF_SILENT;    // do not prompt user
+  file_options.fFlags = FOF_NOCONFIRMATION | FOF_SILENT;  // do not prompt user
   file_options.fAnyOperationsAborted = FALSE;
   file_options.lpszProgressTitle = nullptr;
   file_options.hNameMappings = nullptr;
 
-  SHFileOperation(&file_options);
+  auto ret = SHFileOperation(&file_options);
   delete[] temp_dir;
+
+  if (0 != ret) {
+    return false;
+  }
 #else
   DIR * dir = opendir(p.string().c_str());
   struct dirent * directory_entry;
@@ -590,7 +594,7 @@ inline bool remove_all(const path & p)
   remove(p);
 #endif
 
-  return rcpputils::fs::exists(p);
+  return !rcpputils::fs::exists(p);
 }
 
 /**
--- a/include/rcpputils/filesystem_helper.hpp
+++ b/include/rcpputils/filesystem_helper.hpp
@@ -70,6 +70,7 @@
 #  include <io.h>
 #  define access _access_s
 #else
+#  include <dirent.h>
 #  include <sys/types.h>
 #  include <unistd.h>
 #endif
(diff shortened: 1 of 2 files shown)


include/rcpputils/filesystem_helper.hpp
```diff
@@ -536,6 +537,62 @@ inline bool remove(const path & p)
...
+ * \return true if the directory exists and it was successfully removed, false otherwise.
+ */
+inline bool remove_all(const path & p)
+{
+  if (!is_directory(p)) {return false;}
+
+#ifdef _WIN32
+  // We need a string of type PCZZTSTR, which is a double null terminated char ptr
+  size_t length = p.string().size();
+  TCHAR * temp_dir = new TCHAR[length + 2];
```
ivanpauno, line 555: shouldn't be a `PCZZTSTR` instead of a `TCHAR *`?

include/rcpputils/filesystem_helper.hpp
```diff
@@ -536,6 +537,62 @@ inline bool remove(const path & p)
...
+  file_options.hwnd = nullptr;
+  file_options.wFunc = FO_DELETE;    // delete (recursively)
+  file_options.pFrom = temp_dir;
+  file_options.pTo = nullptr;
+  file_options.fFlags = FOF_NOCONFIRMATION | FOF_SILENT;    // do not prompt user
+  file_options.fAnyOperationsAborted = FALSE;
+  file_options.lpszProgressTitle = nullptr;
+  file_options.hNameMappings = nullptr;
+
+  SHFileOperation(&file_options);
```
ivanpauno, line 570: check return value? check fAnyOperationsAborted status?
See https://docs.microsoft.com/en-us/windows/win32/api/shellapi/nf-shellapi-shfileoperationa#return-value (I guess SHFileOperation is the same that SHFileOperationA :joy:)

include/rcpputils/filesystem_helper.hpp
```diff
@@ -536,6 +537,62 @@ inline bool remove(const path & p)
...
+        return false;
+      }
+    }
+  }
+  closedir(dir);
+  // directory is empty now, call remove
+  remove(p);
+#endif
+
+  return rcpputils::fs::exists(p);
```
ivanpauno, line 593: >  * \return true if the directory exists and it was successfully removed, false otherwise.

should be negated?

test/test_filesystem_helper.cpp
```diff
@@ -302,6 +302,40 @@ TEST(TestFilesystemHelper, filesystem_manipulation)
...
+    ASSERT_TRUE(rcpputils::fs::exists(file));
+    ASSERT_FALSE(rcpputils::fs::remove_all(file));
+    // File still has to be existing
+    ASSERT_TRUE(rcpputils::fs::exists(file));
+  }
+
+  // remove shall fail given that directory is not empty
+  ASSERT_FALSE(rcpputils::fs::remove(dir));
+
+  rcpputils::fs::remove_all(dir);
```
ivanpauno, line 331: check return value

test/test_filesystem_helper.cpp
```diff
@@ -302,6 +302,40 @@ TEST(TestFilesystemHelper, filesystem_manipulation)
...
+
+  for (auto i = 0u; i < num_files; ++i) {
+    std::string file_name = std::string("test_file") + std::to_string(i) + ".txt";
+    auto file = dir / file_name;
+    ASSERT_FALSE(rcpputils::fs::exists(file));
+  }
+  ASSERT_FALSE(rcpputils::fs::exists(dir));
```
ivanpauno, line 338: it would be great to add test for:
* non existent path
* empty dir

include/rcpputils/filesystem_helper.hpp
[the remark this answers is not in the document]
Karsten1987: well, we copy the size of the string, which does not include the null-termination as far as I know.

Karsten1987: > PCZZTSTR may sound scary, but essentially it's just a simple C-style string of TCHARs, which is terminated with two '\0' characters. P (pointer to a), CZ (C-style, zero-terminated) Z (doubly zero-terminated), T (TCHAR), STR (string).

https://stackoverflow.com/a/11010402/4583130

ivanpauno: sure. But if `PCZZTSTR` is equivalent to `TCHAR *` double null terminated, why not using that type?
I don't mind much though.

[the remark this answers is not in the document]
ivanpauno: you're right, thanks for clarifying.

test/test_filesystem_helper.cpp
Karsten1987: see https://github.com/ros2/rcpputils/pull/79/commits/2c08ce03ba174f3c09da7651d70686b89249ac0c

Karsten1987:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11364/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6591/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9316/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11254/)

include/rcpputils/filesystem_helper.hpp
```diff
@@ -547,28 +547,32 @@ inline bool remove(const path & p)
...
   file_options.lpszProgressTitle = nullptr;
   file_options.hNameMappings = nullptr;
 
-  SHFileOperation(&file_options);
+  auto ret = SHFileOperation(&file_options);
   delete[] temp_dir;
+
+  if (0 != ret) {
+    return false;
+  }
```
ivanpauno, line 575:
```suggestion
  return 0 == ret && FALSE == file_options.fAnyOperationsAborted;
```

> It is good practice to examine the value of the fAnyOperationsAborted member of the SHFILEOPSTRUCT. SHFileOperation can return 0 for success if the user cancels the operation. If you do not check fAnyOperationsAborted as well as the return value, you cannot know that the function accomplished the full task you asked of it and you might proceed under incorrect assumptions.

Based on the docs, I would check `fAnyOperationsAborted` (though the user can't cancel the operation in our case).

Karsten1987:
windows once more for verifying the latest changes: [Build Status](https://ci.ros2.org/job/ci_windows/11256/)

dirk-thomas:
This needs to be released into `rolling` in order for http://build.ros2.org/view/Rdev/job/Rdev__rosbag2__ubuntu_focal_amd64/6/ to build.

PR #80, opened 2020-07-01
[foxy backport]: filesystem helpers: adding remove_all to remove non-empty directories…

Karsten1987:
foxy backport of #79

### DIFF
--- a/include/rcpputils/filesystem_helper.hpp
+++ b/include/rcpputils/filesystem_helper.hpp
@@ -48,6 +48,7 @@
 #include <sys/stat.h>
 
 #include <algorithm>
+#include <cstring>
 #include <string>
 #include <vector>
 
--- a/include/rcpputils/filesystem_helper.hpp
+++ b/include/rcpputils/filesystem_helper.hpp
@@ -70,6 +70,7 @@
 #  include <io.h>
 #  define access _access_s
 #else
+#  include <dirent.h>
 #  include <sys/types.h>
 #  include <unistd.h>
 #endif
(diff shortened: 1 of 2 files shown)


ivanpauno:
should features be backported?
Except this is blocking another backport, I don't see much on adding this to Foxy.

emersonknapp:
I am also confused about backporting features into LTS, I thought it was a no-no. Maybe it's just that we have to maintain API and ABI compatibility - if so how to we verify that that is the case?

Karsten1987:
my thinking here was given that it's only additions, it's probably fine to backport. I was planning on using it for backports on rosbag2. But I am okay closing this PR if it's the general consensus.

clalancette:
> I am also confused about backporting features into LTS, I thought it was a no-no. Maybe it's just that we have to maintain API and ABI compatibility - if so how to we verify that that is the case?

Right, API and ABI compatibility are absolutely essential.  Beyond that, whether to backport a feature to a stable release comes down to how broadly applicable the feature is and the maintainers discretion.

Checking for API compatibility is easy; just look at the diff and see if any of the public APIs have been changed.  Checking for ABI compatibility can be a bit more tricky.  Basically anything that changes the size of an object is going to break ABI.  This usually includes adding virtual methods, adding member variables, and the like.  There are some tools around this; one of them is called [abi-compliance-checker](https://lvc.github.io/abi-compliance-checker/), and can be installed on Ubuntu with `sudo apt-get install abi-compliance-checker`.  The buildfarm can also be configured to check ABI, by setting `test_abi: true`, like [the rcl entry](https://github.com/ros/rosdistro/blob/8d3e848f92088090212bcec8a8d40ef6734f95c0/foxy/distribution.yaml#L1369).

emersonknapp:
Well, this doesn't touch any classes/objects or existing APIs. So, as just an addition of a new function it seems it's API and ABI compatible, so I have no problem with backporting

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11427/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6642/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9369/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11330/)

PR #81, opened 2020-07-01
Add cstring include for strcmp

allenh1:
Getting a build failure from source because of a call to `strcmp`, which is in header `cstring` (or `string.h`).

https://github.com/ros2/rcpputils/blob/a2e52e01fb93054d6bee51d0495917deaa6aa7f4/include/rcpputils/filesystem_helper.hpp#L579

Signed-off-by: Hunter L. Allen <hunter.allen@ghostrobotics.io>

### DIFF
--- a/include/rcpputils/filesystem_helper.hpp
+++ b/include/rcpputils/filesystem_helper.hpp
@@ -48,6 +48,7 @@
 #include <sys/stat.h>
 
 #include <algorithm>
+#include <cstring>
 #include <string>
 #include <vector>
 


prajakta-gokhale:
I saw the same failure in ros-tooling/libstatistics_collector running on Ubuntu Focal: [test run](https://github.com/ros-tooling/libstatistics_collector/runs/828543389?check_suite_focus=true#step:4:7207)

Karsten1987:
sorry for the inconvenience.

allenh1:
> sorry for the inconvenience.

Not a problem, was a pretty simple fix. ;)

>  Wondering though why CI didn't pick this up. What platform are you building on?

@Karsten1987 I was also wondering that... This is a source install on Ubuntu 20.04, but using GCC 8 (maybe that's why?).

PR #85, opened 2020-07-10
Add clamp header

### LINKED ISSUE #84: Adding common math functions
This issue comes from the an issue during ROS2 Control development: https://github.com/ros-controls/ros2_controllers/issues/40

There are some fairly common math operations (in this case [clamp](https://en.cppreference.com/w/cpp/algorithm/clamp)) that are not accessible unless you use C++17 or Boost.

Do you believe these operations have a place in rcpputils? Should I make a `clamp.hpp` header just for it?

v-lopez:
Avoids temporarily avoids using boost or C++17

In the end I did not need to create another library target for this, since all templates. But I've put in the separate directory as discussed.

Fixes #84

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -115,6 +115,8 @@ if(BUILD_TESTING)
   set_tests_properties(test_find_library PROPERTIES
     ENVIRONMENT
       "_TEST_LIBRARY_DIR=$<TARGET_FILE_DIR:test_library>;_TEST_LIBRARY=$<TARGET_FILE:test_library>")
+
+  ament_add_gtest(test_clamp test/test_clamp.cpp)
 endif()
 
 ament_package()
--- a/include/rcppmath/clamp.hpp
+++ b/include/rcppmath/clamp.hpp
@@ -1,0 +1,46 @@
+// Copyright 2020 PAL Robotics S.L.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/*! \file clamp.hpp
+ * \brief Restrict a value between two bounds
+ */
+
+#ifndef RCPPMATH__CLAMP_HPP_
+#define RCPPMATH__CLAMP_HPP_
+
+#include <cassert>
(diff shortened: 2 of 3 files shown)


Karsten1987:
This looks good to me, but I guess it deserves a bit more doxygen. It would be great if you could come up with something similar to what is done in other headers, e.g.: https://github.com/ros2/rcpputils/blob/master/include/rcpputils/join.hpp#L30-L39

Karsten1987:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11413/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6630/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9357/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11316/)

emersonknapp:
That is my only feedback as well, to more thoroughly document the functions provided. Thanks for this contribution!

v-lopez:
Sure, I'll do it tomorrow morning.

emersonknapp:
Also note that you will need to use `git commit --signoff` to pass the DCO check. You can do this for existing commits with `git rebase --signoff -i origin/master` and leaving all the commits as `pick`

v-lopez:
I've addressed the documentation and signing, please take a look when you have time.

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11443/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6657/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9386/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11345/)

Karsten1987:
@emersonknapp can this be merged? Ideally we'd like to have this backported to Foxy as well to have ros_control still be compatible with Foxy.

wjwwood:
Uh, I just saw this in the `include` folder. What's the rationale for putting this in a different namespace from `rcpputils`?

We have a pretty consistent pattern of having the root cpp namespaces (and include namespace to match) be the same as the package name.

In my opinion, this should be moved into the `rcpputils` namespace.

v-lopez:
It was done in a separate namespace after discussion in #84, in case in the future similar functions are added and this ends up on a separate package or structure.

wjwwood:
I see, thanks for the pointer @v-lopez, but I have to really disagree with the direction. I think it should be in the `rcpputils` namespace until it is split into a separate package. Splitting it preemptively is just confusing to me. 

I'll probably make pull requests to move it in rolling and deprecate it and provide the new location in foxy, unless someone convinces me otherwise. @emersonknapp @clalancette?

clalancette:
> I'll probably make pull requests to move it in rolling and deprecate it and provide the new location in foxy, unless someone convinces me otherwise. @emersonknapp @clalancette?

That sounds good to me.

PR #92, opened 2020-08-03
[WIP] Add mimick based patch() functionality.

hidmic:
Spin off from https://github.com/ros2/rcutils/pull/272. Still lacking documentation, opening for early review.

### DIFF
--- a/include/rcpputils/patch.hpp
+++ b/include/rcpputils/patch.hpp
@@ -1,0 +1,200 @@
+// Copyright 2020 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RCPPUTILS__PATCH_HPP_
+#define RCPPUTILS__PATCH_HPP_
+
+#include <functional>
+#include <string>
+#include <utility>
+
+#include "mimick/mimick.h"
+#include "rcutils/macros.h"
+
+namespace rcpputils
+{
+
+template<size_t N, typename SignatureT>
+struct patch_traits;
+
+template<size_t N, typename ReturnType>
+struct patch_traits<N, ReturnType(void)>
+{
+  mmk_mock_define(mock_type, ReturnType);
+};
+
+template<size_t N, typename ReturnType, typename ArgType0>
+struct patch_traits<N, ReturnType(ArgType0)>
+{
+  mmk_mock_define(mock_type, ReturnType, ArgType0);
+};
+
+template<size_t N, typename ReturnType,
+  typename ArgType0, typename ArgType1>
+struct patch_traits<N, ReturnType(ArgType0, ArgType1)>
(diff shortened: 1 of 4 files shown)


test/test_patch.cpp
```diff
@@ -0,0 +1,40 @@
...
+RCPPUTILS_PATCH_IGNORE_OPERATOR(rcutils_allocator_t, >)
+
+TEST(TestPatch, test_patched_get_executable_name) {
+  {
+    auto p = rcpputils::patch(
+      "self", rcutils_get_executable_name,
+      [](rcutils_allocator_t allocator) {
+        return rcutils_strdup("dummy_name", allocator);
+      });
+    EXPECT_EQ("dummy_name", rcpputils::get_executable_name());
```
brawner, line 36: EXPECT_STREQ or EXPECT_EQ?

include/rcpputils/patch.hpp
```diff
@@ -0,0 +1,200 @@
...
+
+template<size_t N, typename SignatureT>
+auto make_patch(const std::string & target, std::function<SignatureT> proxy)
+{
+  return patch<N, SignatureT>(target, proxy);
+}
+
+}  // namespace rcpputils
+
+#define RCPPUTILS_PATCH_IGNORE_OPERATOR(type, op) \
```
brawner, line 182: I'm sure you'll add more info in comments, but what purpose does overloading these operators have? It's not clear from the macro name.

include/rcpputils/patch.hpp
```diff
@@ -0,0 +1,200 @@
...
+#include <utility>
+
+#include "mimick/mimick.h"
+#include "rcutils/macros.h"
+
+namespace rcpputils
+{
+
+template<size_t N, typename SignatureT>
+struct patch_traits;
```
brawner, line 29: For these structs, c++ style suggests using PascalCase yah?

include/rcpputils/patch.hpp
```diff
@@ -0,0 +1,200 @@
...
+
+#define RCPPUTILS_PATCH_PROXY(func) \
+  [] (auto && ... args)->decltype(auto) { \
+    return func(std::forward<decltype(args)>(args)...); \
+  }
+
+#define RCPPUTILS_PATCH_TARGET(what, where) \
+  (std::string(RCUTILS_STRINGIFY(where)) + "@" + (what))
+
+#define patch(what, where, with) \
```
brawner, line 195: Your macro name is the same as your struct name, will that lead to confusion?

include/rcpputils/patch.hpp
hidmic: Fair enough. I tend to fallback to `stl` style whenever writing templates.

include/rcpputils/patch.hpp
hidmic: For every argument type, a set of comparison operators must be defined (even if meaningless)  or otherwise `mimick`'s macro generated code won't compile. This macro attempts to be a handy way to generate these dummy overloads. Name is up for debate :sweat_smile: 

FWIW I tried to come up with a smarter mechanism to hide this but couldn't find a way.

test/test_patch.cpp
hidmic: `EXPECT_EQ`, as `rcpputils::get_executable_name()` returns an `std::string`.

include/rcpputils/patch.hpp
hidmic: Perhaps. I don't expect the base class being used directly, but we can rename it.

hidmic:
Based on https://github.com/ros2/rcutils/pull/272#discussion_r463206952, I think the right move here is to push this functionality into our `Mimick` fork and upstream afterwards.

hidmic:
Closing this PR.

PR #95, opened 2020-09-21
Make sure that an existing path is a directory for create_directories

### LINKED ISSUE #94: rcpputils::fs::create_directories doesn't check if an existing path is a directory
Calling `rcpputils::fs::create_directories` on `/a/b/c/d`, where `d` is a file and not a directory, will return `true` (assuming that `/a/b/c/` exists) because it only checks that the path exists.

https://github.com/ros2/rcpputils/blob/5e9849794d1d242322909bf6746779698cb64ce0/include/rcpputils/filesystem_helper.hpp#L506

One could simply call `is_directory` to double check, but looking at the documentation, I don't think that's right:

https://github.com/ros2/rcpputils/blob/5e9849794d1d242322909bf6746779698cb64ce0/include/rcpputils/filesystem_helper.hpp#L493

christophebedard:
If the final/complete path already exists, make sure that it is a directory.

I've also added a line in the relevant test.

Fixes #94

### DIFF
--- a/include/rcpputils/filesystem_helper.hpp
+++ b/include/rcpputils/filesystem_helper.hpp
@@ -511,7 +511,7 @@ inline bool create_directories(const path & p)
 #endif
     }
   }
-  return status == 0;
+  return p_built.is_directory() && status == 0;
 }
 
 /**
--- a/test/test_filesystem_helper.cpp
+++ b/test/test_filesystem_helper.cpp
@@ -286,6 +286,7 @@ TEST(TestFilesystemHelper, filesystem_manipulation)
   EXPECT_TRUE(rcpputils::fs::exists(file));
   EXPECT_TRUE(rcpputils::fs::is_regular_file(file));
   EXPECT_FALSE(rcpputils::fs::is_directory(file));
+  EXPECT_FALSE(rcpputils::fs::create_directories(file));
   EXPECT_GE(rcpputils::fs::file_size(file), expected_file_size);
   EXPECT_THROW(rcpputils::fs::file_size(dir), std::system_error) <<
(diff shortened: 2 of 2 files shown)


wjwwood (approved):
lgtm

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12392/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7360/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10103/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12305/)

christophebedard:
I'll look into that Windows failure.

christophebedard:
Trying something quickly, but if that doesn't work I'll have to actually try on Windows.

* Windows [Build Status](http://ci.ros2.org/job/ci_windows/1​23​15/)

christophebedard:
Not exactly sure why, but it looks like that did the trick. We don't really need to check if the path is a directory at every step/recursively, since the next step will just fail if it's not a directory anyway.

I'll let you decide if you want to re-run CI on all platforms.

wjwwood (approved):
lgtm, new CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12402/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7370/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10113/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12316/)

PR #96, opened 2020-09-24
Revert "Make sure that an existing path is a directory for create_directories"

wjwwood:
Reverts ros2/rcpputils#95

@christophebedard I'm reverting this because it is causing test failures in rosbag, see:

https://ci.ros2.org/job/ci_linux/12453/testReport/

I didn't run CI correctly so we didn't see the break.

@clalancette @jacobperron FYI

PR #97, opened 2020-09-25
create_directories returns true if the path is empty

fujitatomoya:
fix https://github.com/ros2/rcpputils/issues/94, and address https://github.com/ros2/rcpputils/pull/96

### DIFF
--- a/include/rcpputils/filesystem_helper.hpp
+++ b/include/rcpputils/filesystem_helper.hpp
@@ -497,6 +497,10 @@ inline bool create_directories(const path & p)
   path p_built;
   int status = 0;
 
+  if (p.empty()) {
+    return true;
+  }
+
   for (auto it = p.cbegin(); it != p.cend() && status == 0; ++it) {
     if (!p_built.empty() || it->empty()) {
       p_built /= *it;
--- a/include/rcpputils/filesystem_helper.hpp
+++ b/include/rcpputils/filesystem_helper.hpp
@@ -511,7 +511,7 @@ inline bool create_directories(const path & p)
 #endif
     }
   }
-  return status == 0;
+  return p_built.is_directory() && status == 0;
 }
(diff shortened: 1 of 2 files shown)


fujitatomoya:
@wjwwood @christophebedard 

I've added minor fix based on https://github.com/ros2/rcpputils/pull/95.

confirmed rosbag2 test does not detects any failures.

christophebedard:
thank you @fujitatomoya!

clalancette (changes requested):
The problem I have with this change is that this makes us act different than `std::filesystem` (which we eventually want to switch to).  As far as I can tell, if you call:

```
std::filesystem::create_directories("");
```

It throws an exception:
```
terminate called after throwing an instance of 'std::filesystem::__cxx11::filesystem_error'
  what():  filesystem error: cannot create directories: Invalid argument []
```

Also, from a pure semantic point-of-view, the documentation says that it returns true if the directory is created, and false otherwise.  Since it could not create the directory on an empty string, returning true seems wrong here.

My suggestion is that we actually fix the test in rosbag2 that is trying to call this API with an empty string.  Once that is done, we can reapply #95 as it was, and then things shouldn't break on the farm.

christophebedard:
That makes sense. I can do that (@fujitatomoya let me know if you've started working on it).

christophebedard:
> My suggestion is that we actually fix the test in rosbag2 that is trying to call this API with an empty string. Once that is done, we can reapply #95 as it was, and then things shouldn't break on the farm.

I opened https://github.com/ros2/rosbag2/pull/526

fujitatomoya:
@christophebedard @clalancette 

i will close this PR, you can revert https://github.com/ros2/rcpputils/pull/96 after https://github.com/ros2/rosbag2/pull/526.

PR #103, opened 2020-10-08
[Backport Foxy] Add scope_exit helper. (#78)

ahcorde:
This file is required to backport some of the rmw coverage tests to Foxy

Signed-off-by: Michel Hidalgo <michel@ekumenlabs.com>

PR #104, opened 2020-10-09
Update Foxy Github actions

ahcorde:
I had some troubles with the Github Actions on Foxy in particular this PR https://github.com/ros2/rcpputils/pull/103

This should fix the problem

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,10 +17,11 @@ jobs:
       matrix:
           os: [ubuntu-18.04]
     steps:
-    - uses: ros-tooling/setup-ros@0.0.13
-    - uses: ros-tooling/action-ros-ci@0.0.13
+    - uses: ros-tooling/setup-ros@0.0.25
+    - uses: ros-tooling/action-ros-ci@0.0.18
       with:
         package-name: rcpputils
+        target-ros2-distro: foxy
     - uses: actions/upload-artifact@v1
       with:
         name: colcon-logs
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,8 +17,8 @@ jobs:
       matrix:
           os: [ubuntu-18.04]
     steps:
-    - uses: ros-tooling/setup-ros@0.0.25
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -17,10 +17,11 @@ jobs:
       matrix:
           os: [ubuntu-18.04]
     steps:
-    - uses: ros-tooling/setup-ros@0.0.13
-    - uses: ros-tooling/action-ros-ci@0.0.13
+    - uses: ros-tooling/setup-ros@0.0.25
+    - uses: ros-tooling/action-ros-ci@0.0.18
```
clalancette, line 21: I'll suggest that you stick to the same versions as master here for consistency (`setup-ros@0.0.23` and `action-ros-ci@0.0.17`).

If we want to upgrade further, we should probably do it on master first and then backport to Foxy.

clalancette (changes requested):
One thing to fix.

On a separate note, I'm wondering if we should just remove the Github workflow altogether.  We already have PR jobs, and we require CI on all merged PRs anyway, so it seems redundant.  Thoughts?

.github/workflows/test.yml
ahcorde: Updated: https://github.com/ros2/rcpputils/pull/104/commits/1860573f9d93c60cf3e484a5972709337fb442e9

We should probably remove the Github workflow in master (first), right ? and backport the change.

dirk-thomas:
Why does this repo not use the PR CI offered by the ROS buildfarm?

ahcorde:
The ownership of the repo was recently transfer to the Open Robotics https://github.com/ros2/rcpputils/pull/100#event-3830528900. It was managed by AWS, they configure Github Actions for theirs repos.

Anyhow I'm planning to release a new version I can include the CI offered by the ROS buildfarm in Rolling and Foxy

.github/workflows/test.yml
clalancette: > We should probably remove the Github workflow in master (first), right ? and backport the change.

Yeah, that would be my suggestion.

clalancette (approved):

PR #111, opened 2020-11-02
Add an ASSERT to the pointer traits tests.

clalancette:
What's happening here is that we are using the initialization
of the pointer to initialize the "auto" type of the variable,
and then using compile-time checks to check on it.  Because
of this, clang static analysis thinks that the variable is
unused (which it is, at runtime).  Since that is not the
purpose of the test, just put a bogus ASSERT_TRUE in that uses
the variable and quiets clang static analysis.

Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>

### DIFF
--- a/test/test_pointer_traits.cpp
+++ b/test/test_pointer_traits.cpp
@@ -22,6 +22,7 @@
 TEST(TestPointerTraits, is_pointer) {
   auto ptr = new int(13);
   auto ptr_ptr = &ptr;
+  ASSERT_TRUE(ptr_ptr == &ptr);  // to quiet clang static analysis
   const auto * const cptrc = new int(13);
   auto sptr = std::make_shared<int>(13);
   const auto csptr = std::make_shared<int>(13);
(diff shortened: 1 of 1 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12943/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7892/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10660/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12892/)

Karsten1987 (approved):
Might be worth to link to the clang failure for future references.

clalancette:
> Might be worth to link to the clang failure for future references.

I've been running scan-build locally, and slowly picking off the problems.  The exact error it reported is a `Dead initialization`:

```
Value stored to 'ptr_ptr' during its initialization is never read
```

Thanks for the review, I'm going to go ahead and merge this with green CI.

PR #112, opened 2020-11-04
Fix working with filesystem parent paths.

clalancette:
I found two separate bugs while working with parent paths:

1.  Asking for the first parent path of an absolute path
would add a duplicate separator at the beginning.  I fixed
this by checking to see if the separator already exists,
and skiping if there is already one.
2.  Asking for the grandparent path of an absolute path
would return a relative path, which is very incorrect.  The
problem here turned out to be that we were not propagating
empty vector pieces from a child to a parent, and so the
grandparent would miss out that it was supposed to be absolute.
Fix this by always adding all pieces to the parents, even
when they are empty.

I also added tests for the above cases.

Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>

This is currently blocking the resolution of https://github.com/ros/pluginlib/pull/212 .  I'll run CI on everything above rcpputils just to ensure we aren't breaking something.

### DIFF
--- a/include/rcpputils/filesystem_helper.hpp
+++ b/include/rcpputils/filesystem_helper.hpp
@@ -277,10 +277,10 @@ public:
 
     path parent;
     for (auto it = this->cbegin(); it != --this->cend(); ++it) {
-      if (parent.empty() && !this->is_absolute()) {
-        // This handles the case where we are dealing with a relative path;
-        // we don't want a separator at the beginning, so just copy the piece
-        // directly.
+      if (parent.empty() && (!this->is_absolute() || this->is_absolute_with_drive_letter())) {
+        // This handles the case where we are dealing with a relative path or
+        // the Windows drive letter; in both cases we don't want a separator at
+        // the beginning, so just copy the piece directly.
         parent = *it;
       } else {
         parent /= *it;
--- a/include/rcpputils/filesystem_helper.hpp
+++ b/include/rcpputils/filesystem_helper.hpp
@@ -103,10 +103,11 @@ public:
    * \param p A string path split by the platform's string path separator.
    */
   path(const std::string & p)  // NOLINT(runtime/explicit): this is a conversion constructor
-  : path_(p), path_as_vector_(split(p, kPreferredSeparator))
+  : path_(p)
   {
     std::replace(path_.begin(), path_.end(), '\\', kPreferredSeparator);
     std::replace(path_.begin(), path_.end(), '/', kPreferredSeparator);
+    path_as_vector_ = split(path_, kPreferredSeparator);
   }
 
   /**
--- a/include/rcpputils/filesystem_helper.hpp
+++ b/include/rcpputils/filesystem_helper.hpp
@@ -32,9 +32,8 @@
 /*! \file filesystem_helper.hpp
  * \brief Cross-platform filesystem helper functions and additional emulation of [std::filesystem](https://en.cppreference.com/w/cpp/filesystem).
  *
- * If std::filesystem is not available the necessary functions are emulated.
  * Note: Once std::filesystem is supported on all ROS2 platforms, this class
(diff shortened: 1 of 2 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12973/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7923/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10694/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12925/)

include/rcpputils/filesystem_helper.hpp
```diff
@@ -224,7 +223,7 @@ class path
   bool is_absolute() const
   {
     return path_.size() > 0 &&
-           (path_.compare(0, 1, std::string(1, kPreferredSeparator)) == 0 ||
+           (path_[0] == kPreferredSeparator ||
```
fujitatomoya, line 226: just a confirmation, this is only for performance improvement?

fujitatomoya (approved):
lgtm

include/rcpputils/filesystem_helper.hpp
clalancette: Yep, no need to create a new string object to compare one character :).

test/test_filesystem_helper.cpp
```diff
@@ -392,3 +392,15 @@ TEST(TestFilesystemHelper, get_cwd)
   auto p = rcpputils::fs::current_path();
   EXPECT_EQ(expected_dir, p.string());
 }
+
+TEST(TestFilesystemHelper, parent_absolute_path)
+{
+  rcpputils::fs::path path("/home/foo/bar/baz");
+  ASSERT_EQ(path.string(), "/home/foo/bar/baz");
```
mjcarroll, line 399: This new test (and potentially the ones after it) fail Windows CI:

```
[ RUN      ] TestFilesystemHelper.parent_absolute_path
C:\ci\ws\src\ros2\rcpputils\test\test_filesystem_helper.cpp(399): error: Expected equality of these values:
  path.string()
    Which is: "\\home\\foo\\bar\\baz"
  "/home/foo/bar/baz"
[  FAILED  ] TestFilesystemHelper.parent_absolute_path (0 ms)
```

clalancette: It would be Windows, wouldn't it?  Thanks for the heads up, I'll fix it.

clalancette:
All right, I do believe I've fixed the Windows issues.  Kicking off another CI to make sure:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12990/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7942/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10717/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12969/)

clalancette:
@mjcarroll @mjeronimo I could use another review here, thanks.

mjcarroll (approved):
LGTM

clalancette:
All of the CI failures on Windows are known to be flaky, and have failed in the last several Windows builds.

So with green CI everywhere else, and an approval, I'll go ahead and merge.  Thanks for the reviews.

brawner:
It looks like this PR may have introduced an issue on windows debug. It's not immediately obvious to me where this error is occuring though.
https://ci.ros2.org/view/nightly/job/nightly_win_deb/1809/testReport/(root)/projectroot/test_filesystem_helper/

clalancette:
Sigh, yes.  Thanks for pointing it out, I'll take a look.

clalancette:
Windows Debug was right, and I had a bad bug in this patch.  Fix in #113

PR #118, opened 2021-01-11
No windows.h in header files

ivanpauno:
`windows.h` should not be included in any header file because it bizarrely pollutes the global namespace with macros that can easily collide with other names.

My favorites: min, max, ERROR, NO_ERROR,  [and even more](https://stackoverflow.com/questions/2321713/how-do-i-avoid-name-collision-with-macros-defined-in-windows-header-files).

IMO trying to make all our headers compatible with `windows.h` is a bit hopeless, so we should avoid including `windows.h` in header files and use custom workarounds in source files when appropriate (an `#undef` or `#define NOMINMAX` usually fix collisions).

### DIFF
--- a/src/filesystem_helper.cpp
+++ b/src/filesystem_helper.cpp
@@ -67,6 +67,9 @@ namespace rcpputils
 namespace fs
 {
 
+/// \internal Returns true if the path is an absolute path with a drive letter on Windows.
+static bool is_absolute_with_drive_letter(const std::string & path);
+
 path::path(const std::string & p)  // NOLINT(runtime/explicit): this is a conversion constructor
 : path_(p)
 {
@@ -145,7 +148,7 @@ bool path::is_absolute() const
 {
   return path_.size() > 0 &&
           (path_[0] == kPreferredSeparator ||
-          this->is_absolute_with_drive_letter());
+          is_absolute_with_drive_letter(path_));
 }
 
 std::vector<std::string>::const_iterator path::cbegin() const
@@ -170,7 +173,7 @@ path path::parent_path() const
   if (1u == path_as_vector_.size()) {
     if (this->is_absolute()) {
       // Windows is tricky, since an absolute path may start with 'C:\\' or '\\'
-      if (this->is_absolute_with_drive_letter()) {
+      if (is_absolute_with_drive_letter(path_)) {
         return path(path_as_vector_[0] + kPreferredSeparator);
       }
       return path(std::string(1, kPreferredSeparator));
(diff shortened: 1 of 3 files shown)


ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13396/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8317/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11119/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13453/)

src/filesystem_helper.cpp
```diff
@@ -0,0 +1,234 @@
...
+{
+#ifdef _WIN32
+#ifdef UNICODE
+#error "rcpputils::fs does not support Unicode paths"
+#endif
+  TCHAR temp_path[MAX_PATH];
+  DWORD size = GetTempPathA(MAX_PATH, temp_path);
+  if (size > MAX_PATH || size == 0) {
+    std::error_code ec(static_cast<int>(GetLastError()), std::system_category());
+    throw std::system_error(ec, "cannot get temporary directory path");
```
clalancette, line 111: Nit: `#include <system_error>` at the top for these symbols.

src/filesystem_helper.cpp
```diff
@@ -0,0 +1,234 @@
...
+  }
+  return false;
+#else
+  return ::remove(p.string().c_str()) == 0;
+#endif
+}
+
+bool remove_all(const path & p)
+{
+  if (!is_directory(p)) {return rcpputils::fs::remove(p);}
```
clalancette, line 186: Nit: I know you are just moving this code around, but can we make this more like our usual style?

```suggestion
  if (!is_directory(p)) {
    return rcpputils::fs::remove(p);
  }
```

clalancette (commented):
This is fantastic, and exactly what we need.  Thanks for doing this!

I have two nits inline.  Also, I'm wondering if we should just move the implementation of the constructor that takes a `std::string` into the C++ file as well.  It is minor, but putting it there helps on compile times, and just makes it so all of the implementation is in one place.

src/filesystem_helper.cpp
ivanpauno: done in https://github.com/ros2/rcpputils/pull/118/commits/b3280a4643004b2f81f6022b20bd02f733baa077

src/filesystem_helper.cpp
ivanpauno: done in https://github.com/ros2/rcpputils/pull/118/commits/262bef86ab565de506a683d9f71b3e45cdc24789

ivanpauno:
> I'm wondering if we should just move the implementation of the constructor that takes a std::string into the C++ file as well

Moved all method and free function definitions to the source file in https://github.com/ros2/rcpputils/pull/118/commits/82a0901830a43251ba8b2c2c6f8a5368d54c7e54.

clalancette (approved):
Looks good with green CI.

ivanpauno:
CI (build everything, test rcpputils):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13401/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8322/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11127/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13458/)

ivanpauno:
Same jobs also including https://github.com/ros2/rosbag2/pull/600:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13403/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8323/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11128/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13459/)

ivanpauno:
Windows once again, also testing https://github.com/ros2/rcl_logging/pull/71:

* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13460/)

PR #106, opened 2020-10-13
[Backport Foxy] Removed Github Actions (#105)

ahcorde:
Signed-off-by: ahcorde <ahcorde@gmail.com>

PR #122, opened 2021-01-25
Support loading libraries from system library dirs

iluetkeb:
find_library_path will return the plain library name, instead of "". This allows dlopen to try and find the library, rather than throwing an error immediately when the library is not found in LD_LIBRARY_PATH.
This change is important for running with capabilities, because doing so will clean LD_LIBRARY_PATH and would break most of ROS2 if we cannot fall back to the system library search path.
btw, this assumes that the user is either putting directories into the RPATH or otherwise able to add library paths to /etc/ld.so.config. Either of these should work for people who are using setcap and friends.

### DIFF
--- a/src/find_library.cpp
+++ b/src/find_library.cpp
@@ -21,6 +21,8 @@
 #include <sstream>
 #include <string>
 #include <vector>
+#include <iostream>
+
 
 #include "rcutils/filesystem.h"
 #include "rcutils/get_env.h"
@@ -67,7 +69,7 @@ std::string find_library_path(const std::string & library_name)
       return path;
     }
   }
-  return "";
+  return filename;
 }
 
 }  // namespace rcpputils
--- a/src/find_library.cpp
+++ b/src/find_library.cpp
@@ -21,8 +21,6 @@
 #include <sstream>
 #include <string>
 #include <vector>
-#include <iostream>
-
 
 #include "rcutils/filesystem.h"
 #include "rcutils/get_env.h"


iluetkeb:
The lint failures are not my fault... There's something wrong with the set-env command.

emersonknapp:
Is this a backport to Foxy of a fix that already exists on the latest? If so, can you refer `[backport foxy]` and link the PR that made that fix?

> The lint failures are not my fault... There's something wrong with the set-env command.

That's correct - the action versions are out of date on this branch and https://github.com/ros2/rcpputils/pull/106 is supposed to remove them but seems to have gotten stuck

iluetkeb:
> Is this a backport to Foxy of a fix that already exists on the latest?

No, I just made this against foxy because that's what we use.

If you want me to, I can make it against master and then backport.

emersonknapp:
> If you want me to, I can make it against master and then backport.

That's generally the preferred workflow - this way the latest always has bugs fixed, and we can selectively pick things to move back to released distributions.

fujitatomoya (commented):
> rather than throwing an error immediately when the library is not found in LD_LIBRARY_PATH.

i believe that exception happens only if error on accessing environmental variable.

> find_library_path will return the plain library name, instead of "".

IMO, current implementation explicitly return "" when library is NOT found. with this change, it will return either `/libfoobar.so`(found) and `libfoobar.so`(not found). it is still okay to tell found or not. but i think application needs to check the strings if it is found or not. (burden goes to the application anyway)

instead of that, how about adding new function to return library name based on the platform? and this function can be called in `rcpputils::find_library_path`. in application, it will call `rcpputils::find_library_path` and if it returns "", continue to get library name based on platform then dlopen or ...

btw, either way we take, i think it would be nice to update the explanation comments for that.

iluetkeb:
> i believe that exception happens only if error on accessing environmental variable.

That's one error handling block before. Both when the variable cannot be _and_ when the library isn't find in LD_LIBRARY_PATH (which also happens when LD_LIBRARY_PATH is empty), "" is returned, leading to an error in the using function.

> instead of that, how about adding new function to return library name based on the platform?

This would require changes in the upper-level code, while my change doesn't.

In general, I'm somewhat surprised by the complexity of the existing implementation for finding and loading libraries. It largely replicates what is _already_ implemented by dlopen (if you give a relative library name to dlopen, it will already search LD_LIBRARY_PATH). I would prefer to _reduce_ this complexity rather than add to it. My current change is about as simple as it gets.

fujitatomoya:
> which also happens when LD_LIBRARY_PATH is empty

i see, i was also meaning here.

https://github.com/ros2/rcpputils/blob/1c831254b14c2f27a73564fa85ba6e09c65d75d2/src/find_library.cpp#L58

> This would require changes in the upper-level code, while the proposed change doesn't.

correct, but i am not sure if this is the requirement.

> dlopen (if you give a relative library name to dlopen, it will already search LD_LIBRARY_PATH).

agree.

> I would prefer to reduce this complexity rather than add to it. My current change is about as simple as it gets.

i don't have strong opinion here, but i guess updating doc is still better.

iluetkeb:
@fujitatomoya Thanks for the hints, I will take this into account. As suggested by @emersonknapp, I will retarget this to master, first, then backport it.

fujitatomoya:
@iluetkeb thanks for sharing your thoughts 👍

PR #124, opened 2021-01-29
replace separator with preferred separator only if they are different

iuhilnehc-ynos:
Signed-off-by: Chen Lihui <lihui.chen@sony.com>

### DIFF
--- a/src/filesystem_helper.cpp
+++ b/src/filesystem_helper.cpp
@@ -76,7 +76,8 @@ path::path(const std::string & p)  // NOLINT(runtime/explicit): this is a conver
 {
   if (kPreferredSeparator != '\\') {
     std::replace(path_.begin(), path_.end(), '\\', kPreferredSeparator);
-  } else if (kPreferredSeparator != '/') {
+  }
+  if (kPreferredSeparator != '/') {
     std::replace(path_.begin(), path_.end(), '/', kPreferredSeparator);
   }
   path_as_vector_ = split(path_, kPreferredSeparator);
--- a/src/filesystem_helper.cpp
+++ b/src/filesystem_helper.cpp
@@ -74,8 +74,11 @@ static bool is_absolute_with_drive_letter(const std::string & path);
 path::path(const std::string & p)  // NOLINT(runtime/explicit): this is a conversion constructor
 : path_(p)
 {
-  std::replace(path_.begin(), path_.end(), '\\', kPreferredSeparator);
-  std::replace(path_.begin(), path_.end(), '/', kPreferredSeparator);
+  if (kPreferredSeparator != '\\') {
+    std::replace(path_.begin(), path_.end(), '\\', kPreferredSeparator);
+  } else if (kPreferredSeparator != '/') {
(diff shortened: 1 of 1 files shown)


iuhilnehc-ynos:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13548/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8442/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11271/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13613/)

iuhilnehc-ynos:
CI still not ran, so I'll not trigger a new CI for it.

src/filesystem_helper.cpp
```diff
@@ -74,8 +74,11 @@ static bool is_absolute_with_drive_letter(const std::string & path);
 path::path(const std::string & p)  // NOLINT(runtime/explicit): this is a conversion constructor
 : path_(p)
 {
-  std::replace(path_.begin(), path_.end(), '\\', kPreferredSeparator);
-  std::replace(path_.begin(), path_.end(), '/', kPreferredSeparator);
+  if (kPreferredSeparator != '\\') {
```
fujitatomoya, line 77: I am not really convinced to introduce `if statement` here. i believe this is performance but we can expect that would be really small. in the case of performance, we could introduce platform code w/o `if statement` which is better for performance. besides, this `if statement` expects that `kPreferredSeparator` is either '\\' or '/' only, when it needs to support other platform `RCPPUTILS_IMPL_OS_DIRSEP`, it would be complicated for maintenance.

iuhilnehc-ynos: Yeah, I noticed that. Could you review https://github.com/ros2/rcpputils/pull/124/commits/0aa0c82020f8511b54ee54d6b4243f02507b205b ?

iuhilnehc-ynos: Oh, you mean the `if statement` can be omitted. :thinking:
https://github.com/gcc-mirror/gcc/blob/releases/gcc-9/libstdc%2B%2B-v3/include/bits/stl_algo.h#L4407-L4409
```
      for (; __first != __last; ++__first)
	if (*__first == __old_value)
	  *__first = __new_value;
```

fujitatomoya: yeah, i think that we can keep this open see what others think.

clalancette: Backing up for a minute; what's the original purpose of this PR?  That is, what led you to look into the problem in the first place?  That may help inform my answer.

iuhilnehc-ynos: I just thought to decrease the unuseful action as much as possible.
such as the following action in a loop:

- std::replace(path_.begin(), path_.end(), '/', kPreferredSeparator);   // if kPreferredSeparator is '/'

```
for (; __first != __last; ++__first)
  if (*__first == '/')
    *__first = '/';
```

iuhilnehc-ynos: Close it because it's not approved.

PR #126, opened 2021-02-17
Add create_temp_directory filesystem helper

emersonknapp:
Add utility to create formatted temporary directories in a cross-platform way. Useful for test fixtures.

Supports ros2/rosbag2#660

### DIFF
--- a/include/rcpputils/filesystem_helper.hpp
+++ b/include/rcpputils/filesystem_helper.hpp
@@ -293,7 +293,7 @@ RCPPUTILS_PUBLIC path current_path();
  * \brief Create a directory with the given path p.
  *
  * This builds directories recursively and will skip directories if they are already created.
- * \return Return true if the directory is created, false otherwise.
+ * \return Return true if the directory already exists or is created, false otherwise.
  */
 RCPPUTILS_PUBLIC bool create_directories(const path & p);
 
--- a/include/rcpputils/filesystem_helper.hpp
+++ b/include/rcpputils/filesystem_helper.hpp
@@ -270,11 +270,13 @@ RCPPUTILS_PUBLIC path temp_directory_path();
  * if the generated filename exists and keeps generating until a new one is found. This guarantees
  * that there will be no existing files in the returned directory.
  *
- * \param parent The parent path of the directory that will be created
  * \param base_name User-specified portion of the created directory
+ * \param parent The parent path of the directory that will be created
  * \return A path to a newly-created directory with base_name and a 6-character unique suffix
  */
-RCPPUTILS_PUBLIC path create_temp_directory(const path & parent, const std::string & base_name);
+RCPPUTILS_PUBLIC path create_temp_directory(
+  const std::string & base_name,
+  const path & parent = temp_directory_path());
 
 /**
  * \brief Return current working directory.
--- a/include/rcpputils/filesystem_helper.hpp
+++ b/include/rcpputils/filesystem_helper.hpp
@@ -266,17 +266,19 @@ RCPPUTILS_PUBLIC path temp_directory_path();
 /**
  * \brief Construct a uniquely named temporary directory, in "parent", with format base_name.XXXXXX
  *
- * The output is guaranteed to be a newly-created directory, the underlying utilities check
- * if the generated filename exists and keeps generating until a new one is found. This guarantees
- * that there will be no existing files in the returned directory.
+ * The output, if successful, is guaranteed to be a newly-created directory.
+ * The underlying utilities check if the generated path name exists until a one is found.
+ * This guarantees that there will be no existing files in the returned directory.
  *
  * \param base_name User-specified portion of the created directory
  * \param parent The parent path of the directory that will be created
  * \return A path to a newly-created directory with base_name and a 6-character unique suffix
+ *
+ * \throws std::system_error If any OS APIs do not succeed.
  */
 RCPPUTILS_PUBLIC path create_temp_directory(
   const std::string & base_name,
-  const path & parent = temp_directory_path());
+  const path & parent_path = temp_directory_path());
 
 /**
  * \brief Return current working directory.
--- a/include/rcpputils/filesystem_helper.hpp
+++ b/include/rcpputils/filesystem_helper.hpp
@@ -264,10 +264,10 @@ RCPPUTILS_PUBLIC bool exists(const path & path_to_check);
 RCPPUTILS_PUBLIC path temp_directory_path();
 
 /**
- * \brief Construct a uniquely named temporary directory, in "parent", with format base_name.XXXXXX
+ * \brief Construct a uniquely named temporary directory, in "parent", with format base_nameXXXXXX
  *
  * The output, if successful, is guaranteed to be a newly-created directory.
- * The underlying utilities check if the generated path name exists until a one is found.
+ * The underlying implementation keeps generating paths until one that does not exist is found.
  * This guarantees that there will be no existing files in the returned directory.
  *
  * \param base_name User-specified portion of the created directory
--- a/include/rcpputils/filesystem_helper.hpp
+++ b/include/rcpputils/filesystem_helper.hpp
@@ -255,12 +255,20 @@ RCPPUTILS_PUBLIC bool exists(const path & path_to_check);
 /**
  * \brief Get a path to a location in the temporary directory, if it's available.
  *
+ * This does not create any directories.
+ * On Windows, this uses "GetTempPathA"
+ * On non-Windows, this prefers the environment variable TMPDIR, falling back to /tmp
+ *
  * \return A path to a directory for storing temporary files and directories.
  */
 RCPPUTILS_PUBLIC path temp_directory_path();
 
 /**
  * \brief Construct a uniquely named temporary directory, in "parent", with format base_name.XXXXXX
+ *
+ * The output is guaranteed to be a newly-created directory, the underlying utilities check
+ * if the generated filename exists and keeps generating until a new one is found. This guarantees
+ * that there will be no existing files in the returned directory.
  *
  * \param parent The parent path of the directory that will be created
  * \param base_name User-specified portion of the created directory
--- a/include/rcpputils/filesystem_helper.hpp
+++ b/include/rcpputils/filesystem_helper.hpp
@@ -260,6 +260,15 @@ RCPPUTILS_PUBLIC bool exists(const path & path_to_check);
 RCPPUTILS_PUBLIC path temp_directory_path();
 
 /**
+ * \brief Construct a uniquely named temporary directory, in "parent", with format base_name.XXXXXX
+ *
+ * \param parent The parent path of the directory that will be created
+ * \param base_name User-specified portion of the created directory
+ * \return A path to a newly-created directory with base_name and a 6-character unique suffix
+ */
+RCPPUTILS_PUBLIC path create_temp_directory(const path & parent, const std::string & base_name);
+
+/**
  * \brief Return current working directory.
  *
  * \return The current working directory.
--- a/src/filesystem_helper.cpp
+++ b/src/filesystem_helper.cpp
@@ -306,20 +306,38 @@ path temp_directory_path()
   return path(temp_path);
 }
 
-path create_temp_directory(const std::string & base_name, const path & parent)
+path create_temp_directory(const std::string & base_name, const path & parent_path)
 {
-  auto template_path = base_name + ".XXXXXX";
+  const auto template_path = base_name + "XXXXXX";
+  std::string full_template_str = (parent_path / template_path).string();
+  char * mutable_full_template_data = &full_template_str[0];
 
 #ifdef _WIN32
-  _mktemp_s(&template_path[0], template_path.size() + 1);
-  const auto final_path = parent / template;
-  create_directories(final_path);
-  return final_path
+  errno_t ec = _mktemp_s(mutable_full_template_data, full_template_str.size() + 1);
+  if (ec) {
+    throw std::system_error(ec, "could not format the temp directory name template");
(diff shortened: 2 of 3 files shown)


include/rcpputils/filesystem_helper.hpp
```diff
@@ -259,6 +259,15 @@ RCPPUTILS_PUBLIC bool exists(const path & path_to_check);
  */
 RCPPUTILS_PUBLIC path temp_directory_path();
 
+/**
+ * \brief Construct a uniquely named temporary directory, in "parent", with format base_name.XXXXXX
```
Karsten1987, line 263: why wouldn't this use `temp_directory_path` as listed above this function?

Karsten1987 (commented):
maybe there' a good reason for it, but I'd like to see a bit more documentation between the two functions, `temp_directory_path()` and `create_temp_directory()`

include/rcpputils/filesystem_helper.hpp
emersonknapp: At least the way we were using it in rosbag2 - on Linux we were creating these temporary directories within the test directory, rather than within the OS's `/tmp` temporary directory (on Windows is was using the global temp directory). I was thinking that the user _could_ call it with `temp_directory_path()` as the `parent`, but it wouldn't be mandatory. The main functionality of the utility is to format a guaranteed unique name for use, to avoid overlaps with existing tmp dirs, no matter where it is used.

emersonknapp: I've updated the docstrings for both functions to try and make this more clear

Karsten1987 (approved):

src/filesystem_helper.cpp
```diff
@@ -306,6 +306,22 @@ path temp_directory_path()
   return path(temp_path);
 }
 
+path create_temp_directory(const std::string & base_name, const path & parent)
+{
+  auto template_path = base_name + ".XXXXXX";
```
fujitatomoya, line 311: this only guarantees not to return errno `EINVAL`, but `mkdtemp` possibly has errno from `mkdir`. we can refer to man page, it tells that `Also see mkdir(2) for other possible values for errno.` e.g) what if the user application does not have access permission to that parent directory? in that case, `EACCES` will be return with null pointer.

src/filesystem_helper.cpp
```diff
@@ -306,6 +306,22 @@ path temp_directory_path()
...
+
+#ifdef _WIN32
+  _mktemp_s(&template_path[0], template_path.size() + 1);
+  const auto final_path = parent / template;
+  create_directories(final_path);
+  return final_path
+#else
+  auto full_template = (parent / template_path).string();
+  const char * dir_name = mkdtemp(&full_template[0]);
```
fujitatomoya, line 320: i think we need to check pointer if it is NULL or not, and we cannot guarantee this function succeeds always since system call would fail based on user input.

fujitatomoya (changes requested):

src/filesystem_helper.cpp
```diff
@@ -306,6 +306,22 @@ path temp_directory_path()
   return path(temp_path);
 }
 
+path create_temp_directory(const std::string & base_name, const path & parent)
+{
+  auto template_path = base_name + ".XXXXXX";
+
+#ifdef _WIN32
+  _mktemp_s(&template_path[0], template_path.size() + 1);
+  const auto final_path = parent / template;
```
hidmic, line 315: @emersonknapp 
```suggestion
  const auto final_path = parent / template_path;
```
?

include/rcpputils/filesystem_helper.hpp
```diff
@@ -255,10 +255,29 @@ RCPPUTILS_PUBLIC bool exists(const path & path_to_check);
...
+ * if the generated filename exists and keeps generating until a new one is found. This guarantees
+ * that there will be no existing files in the returned directory.
+ *
+ * \param base_name User-specified portion of the created directory
+ * \param parent The parent path of the directory that will be created
+ * \return A path to a newly-created directory with base_name and a 6-character unique suffix
+ */
+RCPPUTILS_PUBLIC path create_temp_directory(
+  const std::string & base_name,
+  const path & parent = temp_directory_path());
```
hidmic, line 279: @emersonknapp nit: consider
```suggestion
  const path & parent_path = temp_directory_path());
```

src/filesystem_helper.cpp
```diff
@@ -306,6 +306,22 @@ path temp_directory_path()
...
+  create_directories(final_path);
```
hidmic, line 316: @emersonknapp consider checking for `create_directories` return status.

src/filesystem_helper.cpp
```diff
@@ -306,6 +306,22 @@ path temp_directory_path()
   return path(temp_path);
 }
 
+path create_temp_directory(const std::string & base_name, const path & parent)
+{
+  auto template_path = base_name + ".XXXXXX";
+
+#ifdef _WIN32
+  _mktemp_s(&template_path[0], template_path.size() + 1);
```
hidmic, line 314: @emersonknapp consider using [`std::string::c_str()`](https://en.cppreference.com/w/cpp/string/basic_string/c_str) instead.

src/filesystem_helper.cpp
hidmic: Also, consider using [`std::string::c_str()`](https://en.cppreference.com/w/cpp/string/basic_string/c_str) instead.

test/test_filesystem_helper.cpp
```diff
@@ -442,3 +442,21 @@ TEST(TestFilesystemHelper, stream_operator)
...
+
+  const auto tmpdir2 = rcpputils::fs::create_temp_directory(basename);
+  EXPECT_TRUE(tmpdir2.exists());
+  EXPECT_TRUE(tmpdir2.is_directory());
+
+  EXPECT_NE(tmpdir1.string(), tmpdir2.string());
+
+  EXPECT_TRUE(rcpputils::fs::remove_all(tmpdir1));
+  EXPECT_TRUE(rcpputils::fs::remove_all(tmpdir2));
+}
```
hidmic, line 462: @emersonknapp it'd be good to test how the function behaves in pathological cases: an empty basename, a basename that already contains the trailing template, a parent path that doesn't exist, and so on.

src/filesystem_helper.cpp
emersonknapp: In these APIs, the input `char *` must be mutable, because the `XXXXXX` is overwritten by the API. `.c_str()` (or `.data()` in C++11) return a `const char *` that can't be used here.

hidmic: Hmm, good point. I'd rather not make assumptions about `std::string` underlying structure either (i.e. using `&var[0]`). I guess we're left with primitive char arrays or `std::vector<char>` (using `std::vector::data`).

hidmic, line 314: @emersonknapp also, why isn't `_mktemp_s` given a full path? How will it check for file existence?

emersonknapp:
@hidmic @fujitatomoya thanks for the detailed review! I think I have addressed all comments - please take another look and let me know what you think.

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/b25b4e53778519d8b0ff390151634105/raw/105f39d36d8d7985109b661ab24d452bcb7419f8/ros2.repos
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/7665
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13685/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8564/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11402/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13756/)

src/filesystem_helper.cpp
```diff
@@ -306,6 +306,40 @@ path temp_directory_path()
   return path(temp_path);
 }
 
+path create_temp_directory(const std::string & base_name, const path & parent_path)
+{
+  const auto template_path = base_name + "XXXXXX";
+  std::string full_template_str = (parent_path / template_path).string();
+  char * mutable_full_template_data = &full_template_str[0];
```
hidmic, line 313: The problem with `char * mutable_full_template_data = &full_template_str[0];` is the assumption it makes about how `std::string` stores its data. Something like:
```suggestion
  std::vector<char> mutable_full_template_data(full_template_str.begin(), full_template_str.end());
  // now use mutable_full_template_data.data()
```
doesn't have that issue.

clalancette: I don't particularly care which way we go, but I will point out that the [documentation](https://en.cppreference.com/w/cpp/string/basic_string) for string says:

```
The elements of a basic_string are stored contiguously, that is, for a basic_string s, &*(s.begin() + n) == &*s.begin() + n for any n in [0, s.size()), or, equivalently, a pointer to s[0] can be passed to functions that expect a pointer to the first element of a null-terminated (since C++11)CharT[] array.
```

hidmic: Uh, I stand corrected then. Thanks @clalancette. Not sure why I was convinced the standard did not enforce contiguity.

src/filesystem_helper.cpp
hidmic: As mentioned below by @clalancette, `std::basic_string` does guarantee contiguous storage such that `&var[0]` is well-defined. Nevermind my last comment.

include/rcpputils/filesystem_helper.hpp
```diff
@@ -255,10 +255,31 @@ RCPPUTILS_PUBLIC bool exists(const path & path_to_check);
...
+ *
  * \return A path to a directory for storing temporary files and directories.
  */
 RCPPUTILS_PUBLIC path temp_directory_path();
 
+/**
+ * \brief Construct a uniquely named temporary directory, in "parent", with format base_name.XXXXXX
+ *
+ * The output, if successful, is guaranteed to be a newly-created directory.
+ * The underlying utilities check if the generated path name exists until a one is found.
```
hidmic, line 270: @emersonknapp did you mean:
```suggestion
 * The underlying implementation keeps generating paths until one that does not exists is found.
```

src/filesystem_helper.cpp
```diff
@@ -306,6 +306,40 @@ path temp_directory_path()
   return path(temp_path);
 }
 
+path create_temp_directory(const std::string & base_name, const path & parent_path)
+{
+  const auto template_path = base_name + "XXXXXX";
```
hidmic, line 311: @emersonknapp 
```suggestion
  const auto template_path = base_name + ".XXXXXX";
```
? Or otherwise, the docblock needs a change.

test/test_filesystem_helper.cpp
```diff
@@ -442,3 +442,59 @@ TEST(TestFilesystemHelper, stream_operator)
...
+    EXPECT_TRUE(rcpputils::fs::remove_all(tmpdir));
+  }
+
+  // edge case inputs
+  {
+    // Provided no base name we should still get a path with the 6 unique template chars
+    const auto tmpdir_emptybase = rcpputils::fs::create_temp_directory("");
+    EXPECT_EQ(tmpdir_emptybase.filename().string().size(), 6u);
+
+    // Empty path doesn't exist and cannott be created
```
hidmic, line 490: @emersonknapp nit: 
```suggestion
    // Empty path doesn't exist and cannot be created
```

hidmic (approved):
Overall LGTM pending green CI

src/filesystem_helper.cpp
emersonknapp: changing the docblock - I don't want to force the `.`

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13688/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8567/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11405/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13759/)

src/filesystem_helper.cpp
```diff
@@ -306,6 +306,44 @@ path temp_directory_path()
...
+  std::string full_template_str = (parent_path / template_path).string();
+  std::vector<char> mutable_full_template_data(full_template_str.begin(), full_template_str.end());
```
hidmic, line 313: BTW your original code was OK too, as @clalancette pointed out in https://github.com/ros2/rcpputils/pull/126#discussion_r577930719. 

My apologies for the unnecessary noise :facepalm:

src/filesystem_helper.cpp
emersonknapp: Ok - I'll change it back, for performance reasons. This implementation has to perform two full copies that aren't necessary if we just go the `&string[0]` route

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13689/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8568/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11406/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13760/)

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13691/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8570/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11408/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13763/)

fujitatomoya (approved):
lgtm

PR #128, opened 2021-03-22
Make library attributes visible

nnmm:
This change was requested in https://github.com/ros2/rclcpp/pull/1452#discussion_r589166427: have a central place for these attributes instead of duplicating them in rclcpp.

Signed-off-by: Nikolai Morin <nikolai.morin@apex.ai>

### DIFF
--- a/include/rcpputils/find_library.hpp
+++ b/include/rcpputils/find_library.hpp
@@ -26,23 +26,6 @@
 namespace rcpputils
 {
 
-#ifdef _WIN32
-static constexpr char kPathVar[] = "PATH";
-static constexpr char kPathSeparator = ';';
-static constexpr char kSolibPrefix[] = "";
-static constexpr char kSolibExtension[] = ".dll";
-#elif __APPLE__
-static constexpr char kPathVar[] = "DYLD_LIBRARY_PATH";
-static constexpr char kPathSeparator = ':';
-static constexpr char kSolibPrefix[] = "lib";
-static constexpr char kSolibExtension[] = ".dylib";
-#else
-static constexpr char kPathVar[] = "LD_LIBRARY_PATH";
-static constexpr char kPathSeparator = ':';
-static constexpr char kSolibPrefix[] = "lib";
-static constexpr char kSolibExtension[] = ".so";
-#endif
-
 /// Find a library located in the OS's specified environment variable for library paths.
 /**
  *
@@ -57,6 +40,18 @@ static constexpr char kSolibExtension[] = ".so";
  */
 RCPPUTILS_PUBLIC
 std::string find_library_path(const std::string & library_name);
+
+
+/// Create the filename corresponding to the library name.
+/**
+ *
+ * @sa find_library_path() for information about the platform-specific filenames.
+ *
+ * \param[in] library_name Name of the library.
+ * \return The filename for the library.
+ */
+RCPPUTILS_PUBLIC
+std::string filename_for_library(const std::string & library_name);
 
 }  // namespace rcpputils
 
--- a/include/rcpputils/find_library.hpp
+++ b/include/rcpputils/find_library.hpp
@@ -45,7 +45,7 @@ std::string find_library_path(const std::string & library_name);
 /// Create the filename corresponding to the library name.
 /**
  *
- * @sa find_library_path() for information about the platform-specific filenames.
+ * \sa find_library_path() for information about the platform-specific filenames.
  *
  * \param[in] library_name Name of the library.
  * \return The filename for the library.
--- a/include/rcpputils/find_library.hpp
+++ b/include/rcpputils/find_library.hpp
@@ -35,12 +35,22 @@ namespace rcpputils
  *  * Windows: `%PATH%`, `{}.dll`
  *
  * \param[in] library_name Name of the library to find.
- * \return The absolute filesystem path, including the appropriate prefix and extension
+ * \return The absolute filesystem path, including the appropriate prefix and extension, or the
+ * empty string when the library was not found.
  * \throws std::runtime_error if an error is encountered when accessing environment variables.
  */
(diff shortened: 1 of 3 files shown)


nnmm:
@wjwwood Could you review?

wjwwood (commented):
I think it would be better if this were exposed via a function(s), but I don't have a strong reason why.

@clalancette @mjeronimo any feedback (as maintainers)?

emersonknapp (approved):
This seems fine - though my personal preference would be to expose a new function `find_library_in_path` - and keep these implementation details hidden

clalancette:
> This seems fine - though my personal preference would be to expose a new function `find_library_in_path` - and keep these implementation details hidden

Exposing these details seems pretty ugly.  I have to agree with @wjwwood and @emersonknapp that a function would be much nicer here.

nnmm:
@clalancette @emersonknapp `find_library_in_path()` does not do what I'm interested in. What rclcpp_generic does is 
```
auto library_path = package_prefix + dynamic_library_folder + rcpputils::kSolibPrefix +
    package_name + "__" + typesupport_identifier + rcpputils::kSolibExtension;
```
(`dynamic_library_folder` is also platform-specific, but I assume it's a ROS convention that doesn't generalize)

So the only function that we could extract into here is something like `affix_library_name()`, which adds the prefix and the extension. Would you prefer that?

mjeronimo:
It does seem that there is a function or two that wants to come out rather than exposing the platform details. I'm for encapsulating the platform-specifics in the implementation file and providing platform independent functions in the header.

nnmm:
I changed it to expose only a function.

mjeronimo (approved):

include/rcpputils/find_library.hpp
```diff
@@ -41,6 +41,18 @@ namespace rcpputils
 RCPPUTILS_PUBLIC
 std::string find_library_path(const std::string & library_name);
 
+
+/// Create the filename corresponding to the library name.
+/**
+ *
+ * @sa find_library_path() for information about the platform-specific filenames.
```
wjwwood, line 48: Please use \ instead of @

clalancette (approved):

clalancette (changes requested):
Oh, sorry.  Can you just add one test to cover this?  Should be a simple one.

nnmm:
@clalancette It is covered by the existing test since I modified `find_library_path` to use `filename_for_library`, does that count? Otherwise it's a bit tricky – for an exact test, I'd need to know which platforms this test runs on to construct an expected output, and the test would just be a reimplementation of the function.

src/find_library.cpp
```diff
@@ -70,4 +69,9 @@ std::string find_library_path(const std::string & library_name)
   return "";
 }
 
+std::string filename_for_library(const std::string & library_name)
```
emersonknapp, line 72: My personal preference for this was that it would look like 

```
std::string find_library_in_path(const std::string & library_name, const std::string & in_path)
{
  auto name = filename_for_library(library_name);
  auto final_path = in_path / name;
  if (final_path.is_file) {
    return final_path;
  }
  return "";
}
```

To be similar to `find_library_path` - but a more limited version for a _specific_ directory.

nnmm: That's fine with me as well. Can it throw an exception instead of returning an empty string?

emersonknapp: `find_library_path` doesn't throw an exception - I would think the two should act the same. I don't have a strong opinion which way that is, just that they should work the same

clalancette:
> @clalancette It is covered by the existing test since I modified `find_library_path` to use `filename_for_library`, does that count?

Oh yeah, that counts.  OK, fine by me.  I'll add my approval, but you should take care of @emersonknapp 's comments as well.

clalancette (approved):

src/find_library.cpp
nnmm: @emersonknapp Please take another look if this is fine for you. I named it `path_for_library` instead since it doesn't do any finding.

Took me a while to figure out that cppcheck has issues with Gtests in namespaces.

emersonknapp (approved):

test/test_find_library.cpp
```diff
@@ -59,15 +65,26 @@ TEST(test_find_library, find_library)
...
-}  // namespace
-}  // namespace rcpputils
+TEST(test_find_library, library_path)
+{
+  // Get ground-truth values from CTest properties.
+  const auto pair = test_lib_path_and_dir();
+  const std::string expected_library_path = pair.first;
+  const std::string test_library_path = pair.second;
+
+  const std::string test_lib_actual = rcpputils::path_for_library(test_lib_dir, "test_library");
```
emersonknapp, line 85: This is failing to build because `test_lib_dir` is not defined.

nnmm: Ah, sorry about that. It should be fixed now.

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/d8fce7350957a70f43a565b4f7f62c21/raw/b486b71a66a8779f7512fbdf779c11b71c4c5544/ros2.repos
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/7987
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14083/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8901/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11776/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14176/)

nnmm:
@emersonknapp I assume that means it's good to merge?

wjwwood:
@emersonknapp I'm gonna merge this so we can move the other pr forward. 👍

PR #133, opened 2021-06-18
migrate rolling mean from ros2_controllers to rcppmath

Karsten1987:
that's a clear copy from ros2_controllers package to rcppmath. This makes more sense so we can leverage this header across multiple packages.

Ideally, I plan to also backport this to galactic and foxy.
fyi @v-lopez

### DIFF
--- a/include/rcppmath/rolling_mean_accumulator.hpp
+++ b/include/rcppmath/rolling_mean_accumulator.hpp
@@ -1,0 +1,68 @@
+// Copyright 2020 PAL Robotics S.L.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


Karsten1987:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14781/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9530/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12447/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14928/)

emersonknapp:
I think this is also the same thing as https://github.com/ros-tooling/libstatistics_collector/blob/master/include/libstatistics_collector/moving_average_statistics/moving_average.hpp

fujitatomoya (commented):
looks good to me, since we already have rcppmath. is this actually used on somewhere? i am just not familiar with ros2 control packages.

clalancette (approved):

bmagyar:
Can this be added to Galactic please? Or was only meant for after Galactic?  @clalancette @fujitatomoya

aprotyas:
This doesn't break ABI so a backport should be possible, I think.

clalancette:
@mergifyio backport galactic

PR #138, opened 2021-07-20
Improve `rcppmath` Doxygen documentation

aprotyas:
This PR broadly attempts to make `rcppmath` documentation more detailed, as well as fixes some Doxygen issues that previously prevented `rcppmath` documentation from building correctly using `rosdoc2`.

Specifically:
* Documented rolling mean accumulator methods.
* For the clamping methods:
  * Added `\sa` tag instead of explicitly asking user to "see the other function".
  * nit: Capitalization.

Signed-off-by: Abrar Rahman Protyasha <abrar@openrobotics.org>

### DIFF
--- a/include/rcppmath/rolling_mean_accumulator.hpp
+++ b/include/rcppmath/rolling_mean_accumulator.hpp
@@ -26,10 +26,10 @@ namespace rcppmath
 {
 
 /**
- * \brief Simplification of boost::accumulators::accumulator_set<double,
- *  bacc::stats<bacc::tag::rolling_mean>> to avoid dragging boost dependencies.
+ * \brief Computes the mean of the last accumulated elements.
  *
- * Computes the mean of the last accumulated elements.
+ * This is a simplified version of boost's rolling mean accumulator,
+ * written to avoid dragging boost dependencies.
  */
 template<typename T>
 class RollingMeanAccumulator
--- a/include/rcppmath/rolling_mean_accumulator.hpp
+++ b/include/rcppmath/rolling_mean_accumulator.hpp
@@ -29,7 +29,7 @@ namespace rcppmath
  * \brief Computes the mean of the last accumulated elements.
  *
  * This is a simplified version of boost's rolling mean accumulator,
- * written to avoid dragging boost dependencies.
+ * written to avoid dragging in boost dependencies.
  */
 template<typename T>
 class RollingMeanAccumulator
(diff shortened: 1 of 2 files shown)


include/rcppmath/rolling_mean_accumulator.hpp
```diff
@@ -27,20 +27,30 @@ namespace rcppmath
 
 /**
  * \brief Simplification of boost::accumulators::accumulator_set<double,
- *  bacc::stats<bacc::tag::rolling_mean>> to avoid dragging boost dependencies
+ *  bacc::stats<bacc::tag::rolling_mean>> to avoid dragging boost dependencies.
```
clalancette, line 30: I know you didn't add this, but I'd honestly just remove this sentence.  This is an implementation detail, and doesn't tell us what the class is meant to do.

If we really want to keep it, then I'd lead with "Computes the mean of the last accumulated elements", and then somewhere in the extended description mention that it is a simplified version of boost rolling_mean.

clalancette (changes requested):
I've got one thing to change here.  Otherwise it looks good to me.

include/rcppmath/rolling_mean_accumulator.hpp
aprotyas: I concur that this line should be changed. I'll remove the lengthy signature and mention boost rolling_mean in the extended description then.

aprotyas: Fixed in 6cd9a07.

include/rcppmath/rolling_mean_accumulator.hpp
```diff
@@ -26,21 +26,31 @@ namespace rcppmath
 {
 
 /**
- * \brief Simplification of boost::accumulators::accumulator_set<double,
- *  bacc::stats<bacc::tag::rolling_mean>> to avoid dragging boost dependencies
+ * \brief Computes the mean of the last accumulated elements.
  *
- * Computes the mean of the last accumulated elements
+ * This is a simplified version of boost's rolling mean accumulator,
+ * written to avoid dragging boost dependencies.
```
clalancette, line 32:
```suggestion
 * written to avoid dragging in boost dependencies.
```

clalancette (approved):
One minor fixup to do.  I'll approve anyway, assuming you'll fix before merging.

ros-discourse:
This pull request has been mentioned on **ROS Discourse**. There might be relevant details there:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2021-9-16/22372/1

PR #139, opened 2021-08-02
Fix API documentation for clean `rosdoc2` build

aprotyas:
This PR introduces changes which resolve multiple errors/warnings on `rosdoc2 build` invocations to generate API documentation for `rcpputils`.

Namely:
- [x] Doxygen configuration:
  - Inclusion of all required files (including markdown files).
  - Generation of XML tags.
- [x] `rcpputils::fs`: Missing parameters in function header.
- [x] `rcpputils::shared_library`: Remove `@copydoc` directive.
- [x] `rcpputils::split`: Conditionally exclude [`split` signature with template parameter list](https://github.com/ros2/rcpputils/compare/master...aprotyas:aprotyas/fix_api_docs?expand=1#diff-ebc015817b8762a6dad244f0d9f1ad909ec5c5d2ec8eaf679cf29c23b8fd9112R57-R65) since `breathe` has a tough time parsing said list. Once https://github.com/michaeljones/breathe/pull/711 is merged, the `@cond/@endcond` directives can be removed.
- [x] `rcpputils::thread_safety_annotations`: There seems to be no straightforward method to address the "anonymous hyperlink" warning generated because of the double trailing underscores in the `RCPPUTILS_THREAD_ANNOTATION_ATTRIBUTE__` macro. As such, this macro has been conditionally excluded.

### DIFF
--- a/include/rcpputils/filesystem_helper.hpp
+++ b/include/rcpputils/filesystem_helper.hpp
@@ -83,7 +83,7 @@ public:
   /**
    * \brief Conversion constructor from a std::string path.
    *
-   * \param p A string path split by the platform's string path separator.
+   * \param[in] p A string path split by the platform's string path separator.
    */
   RCPPUTILS_PUBLIC
   path(const std::string & p);  // NOLINT(runtime/explicit): this is a conversion constructor
@@ -183,7 +183,7 @@ public:
   /**
   * \brief Concatenate a path and a string into a single path.
   *
-  * \param other the string compnoent to concatenate
+  * \param[in] other the string compnoent to concatenate
   * \return The combined path of this and other.
   */
   RCPPUTILS_PUBLIC path operator/(const std::string & other) const;
@@ -191,7 +191,7 @@ public:
   /**
   * \brief Append a string component to this path.
   *
-  * \param other the string component to append
+  * \param[in] other the string component to append
   * \return *this
   */
   RCPPUTILS_PUBLIC path & operator/=(const std::string & other);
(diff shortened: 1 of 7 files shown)


aprotyas:
Not running CI since all changes introduced in this PR are for documentation.

include/rcpputils/filesystem_helper.hpp
```diff
@@ -271,7 +271,7 @@ RCPPUTILS_PUBLIC path temp_directory_path();
  * This guarantees that there will be no existing files in the returned directory.
  *
  * \param base_name User-specified portion of the created directory
- * \param parent The parent path of the directory that will be created
+ * \param parent_path The parent path of the directory that will be created
```
clalancette, line 274: Can I suggest that we change all of the parameter lines in this file to be `\param[in]`, `\param[out]`, or `\param[inout]`?  That will be a bigger patch, but will be more consistent with what we use elsewhere in this package.

clalancette (changes requested):
One slightly large thing to change, then this will be ready to go.

include/rcpputils/filesystem_helper.hpp
aprotyas: Yes, that makes sense. Addressed in 22eeba1. `rcpputils/assert.hpp` received the same treatment for consistency.

clalancette (approved):

ros-discourse:
This pull request has been mentioned on **ROS Discourse**. There might be relevant details there:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2021-9-16/22372/1

PR #145, opened 2021-09-16
Add new interface for time conversion

Barry-Xu-2018:
Add a new interface `rcpputils::rcutils_duration_cast`.   

About this new interface, it is mentioned at ros2/rclcpp#1662 (https://github.com/ros2/rclcpp/pull/1662#discussion_r707634117)

### DIFF
--- a/include/rcpputils/time.hpp
+++ b/include/rcpputils/time.hpp
@@ -37,9 +37,9 @@ std::chrono::nanoseconds convert_to_nanoseconds(
   const std::chrono::duration<DurationRepT, DurationT> & time)
 {
   // Casting to a double representation might lose precision and allow the check below to succeed
-  // but the actual cast to nanoseconds fail. Using 1 DurationT worth of nanoseconds less than max
+  // but the actual cast to nanoseconds fail. Using 1 worth of nanoseconds less than max
   constexpr auto maximum_safe_cast_ns =
-    std::chrono::nanoseconds::max() - std::chrono::duration<DurationRepT, DurationT>(1);
+    std::chrono::nanoseconds::max() - std::chrono::nanoseconds(1);
   // If period is greater than nanoseconds::max(), the duration_cast to nanoseconds will overflow
   // a signed integer, which is undefined behavior. Checking whether any std::chrono::duration is
   // greater than nanoseconds::max() is a difficult general problem. This is a more conservative
--- a/include/rcpputils/time.hpp
+++ b/include/rcpputils/time.hpp
@@ -1,0 +1,62 @@
+// Copyright 2021 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RCPPUTILS__TIME_HPP_
+#define RCPPUTILS__TIME_HPP_
+
+#include <chrono>
+
+#include "rcutils/time.h"
+
+#include "rcpputils/visibility_control.hpp"
+
+namespace rcpputils
+{
+
+/// Convert to rcutils duration value.
+/*
+ * \param[in] time The time to be converted to rcutils duration.
+ * \return rcutils duration value
+ * \throws std::invalid_argument if time is bigger than std::chrono::nanoseconds::max().
+ */
+template<typename DurationRepT = int64_t, typename DurationT = std::milli>
+RCPPUTILS_PUBLIC
+rcutils_duration_value_t convert_to_rcutils_duration(
+  const std::chrono::duration<DurationRepT, DurationT> & time)
+{
+  // Casting to a double representation might lose precision and allow the check below to succeed
+  // but the actual cast to nanoseconds fail. Using 1 DurationT worth of nanoseconds less than max
+  constexpr auto maximum_safe_cast_ns =
+    std::chrono::nanoseconds::max() - std::chrono::duration<DurationRepT, DurationT>(1);
+  // If period is greater than nanoseconds::max(), the duration_cast to nanoseconds will overflow
+  // a signed integer, which is undefined behavior. Checking whether any std::chrono::duration is
+  // greater than nanoseconds::max() is a difficult general problem. This is a more conservative
+  // version of Howard Hinnant's (the <chrono> guy>) response here:
+  // https://stackoverflow.com/a/44637334/2089061
+  // However, this doesn't solve the issue for all possible duration types of period.
+  // Follow-up issue: https://github.com/ros2/rclcpp/issues/1177
+  constexpr auto ns_max_as_double =
+    std::chrono::duration_cast<std::chrono::duration<double, std::chrono::nanoseconds::period>>(
+    maximum_safe_cast_ns);
+  if (time > ns_max_as_double) {
+    throw std::invalid_argument{
+            "time must be less than std::chrono::nanoseconds::max()"};
+  }
+
+  return (std::chrono::duration_cast<std::chrono::nanoseconds>(time)).count();
+}
+
+}  // namespace rcpputils
+
+#endif  // RCPPUTILS__TIME_HPP_
--- a/include/rcpputils/time.hpp
+++ b/include/rcpputils/time.hpp
@@ -19,8 +19,6 @@
 
 #include "rcutils/time.h"
 
-#include "rcpputils/visibility_control.hpp"
-
 namespace rcpputils
 {
 
@@ -34,7 +32,6 @@ namespace rcpputils
  * \throws std::invalid_argument if time is bigger than std::chrono::nanoseconds::max().
  */
 template<typename DurationRepT = int64_t, typename DurationT = std::milli>
-RCPPUTILS_PUBLIC
 std::chrono::nanoseconds convert_to_nanoseconds(
   const std::chrono::duration<DurationRepT, DurationT> & time)
 {
--- a/include/rcpputils/time.hpp
+++ b/include/rcpputils/time.hpp
@@ -32,25 +32,13 @@ namespace rcpputils
  * \throws std::invalid_argument if time is bigger than std::chrono::nanoseconds::max() or less than
  * std::chrono::nanoseconds::min().
  */
-template<typename DurationRepT = int64_t, typename DurationT = std::milli>
+template<typename DurationRepT, typename DurationT>
 std::chrono::nanoseconds convert_to_nanoseconds(
   const std::chrono::duration<DurationRepT, DurationT> & time)
 {
-  // Casting to a double representation might lose precision and allow the check below to succeed
-  // but the actual cast to nanoseconds fail. Using 1 worth of nanoseconds less than max
-  constexpr auto maximum_safe_cast_ns =
-    std::chrono::nanoseconds::max() - std::chrono::nanoseconds(1);
-  // If period is greater than nanoseconds::max(), the duration_cast to nanoseconds will overflow
-  // a signed integer, which is undefined behavior. Checking whether any std::chrono::duration is
-  // greater than nanoseconds::max() is a difficult general problem. This is a more conservative
-  // version of Howard Hinnant's (the <chrono> guy>) response here:
-  // https://stackoverflow.com/a/44637334/2089061
-  // However, this doesn't solve the issue for all possible duration types of period.
-  // e.g std::chrono::hours(10000000) cannot be converted to nanoseconds.
-  // Follow-up issue: https://github.com/ros2/rclcpp/issues/1177
   constexpr auto ns_max_as_double =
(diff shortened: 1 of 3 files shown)


include/rcpputils/time.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+/// Convert to rcutils duration value.
+/*
+ * \param[in] time The time to be converted to rcutils duration.
+ * \return rcutils duration value
+ * \throws std::invalid_argument if time is bigger than std::chrono::nanoseconds::max().
+ */
+template<typename DurationRepT = int64_t, typename DurationT = std::milli>
+RCPPUTILS_PUBLIC
+rcutils_duration_value_t convert_to_rcutils_duration(
```
fujitatomoya, line 35: how about returning as `std::chrono::nanoseconds` instead of `rcutils_duration_value_t` as C++ context?

include/rcpputils/time.hpp
```diff
@@ -0,0 +1,62 @@
...
+  // Follow-up issue: https://github.com/ros2/rclcpp/issues/1177
+  constexpr auto ns_max_as_double =
+    std::chrono::duration_cast<std::chrono::duration<double, std::chrono::nanoseconds::period>>(
+    maximum_safe_cast_ns);
+  if (time > ns_max_as_double) {
+    throw std::invalid_argument{
+            "time must be less than std::chrono::nanoseconds::max()"};
+  }
+
+  return (std::chrono::duration_cast<std::chrono::nanoseconds>(time)).count();
```
fujitatomoya, line 57: we do not need to check `integer overflow`? i am not sure if this will never happens though.

```
  if (period_ns < std::chrono::nanoseconds::zero()) {
    throw std::runtime_error{
            "Casting timer period to nanoseconds resulted in integer overflow."};
  }
```

Barry-Xu-2018:
```c++
  constexpr auto maximum_safe_cast_ns =
    std::chrono::nanoseconds::max() - std::chrono::duration<DurationRepT, DurationT>(1);

  constexpr auto ns_max_as_double =
    std::chrono::duration_cast<std::chrono::duration<double, std::chrono::nanoseconds::period>>(
    maximum_safe_cast_ns);
  if (time > ns_max_as_double) {
    throw std::invalid_argument{
            "time must be less than std::chrono::nanoseconds::max()"};
  }
```
Above code make sure overflow doesn't occur. It gets std::chrono::duration<double, std::chrono::nanoseconds::period> with the maximum value of int64. And compare with `time`. If overflow, exception is throwed.

include/rcpputils/time.hpp
Barry-Xu-2018: Yeah. But if use `std::chrono::nanoseconds`, user needs to call count().  

How about this ?
```c++
template<typename DurationRepT = int64_t, typename DurationT = std::milli>
RCPPUTILS_PUBLIC
int64_t convert_to_nanoseconds(const std::chrono::duration<DurationRepT, DurationT> & time)
```

fujitatomoya: > But if use std::chrono::nanoseconds, user needs to call count().

then user can call `count` method? imo, i think that is okay for user application.

include/rcpputils/time.hpp
fujitatomoya: thanks, original code has this check just in case, so i was just wondering that would be needed.

include/rcpputils/time.hpp
Barry-Xu-2018: Okay.

fujitatomoya (approved):

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15253/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9972/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12937/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15449/)

include/rcpputils/time.hpp
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * This function help to convert from std::chrono::duration to std::chrono::nanoseconds and throw
+ * exception if overflow occurs while coverting.
+ *
+ * \param[in] time The time to be converted to std::chrono::nanoseconds.
+ * \return std::chrono::nanoseconds.
+ * \throws std::invalid_argument if time is bigger than std::chrono::nanoseconds::max().
+ */
+template<typename DurationRepT = int64_t, typename DurationT = std::milli>
+RCPPUTILS_PUBLIC
```
aprotyas, line 37: `dllimport` is only required at declaration, not definition. Removing this line should fix Windows CI.

test/test_time.cpp
```diff
@@ -0,0 +1,28 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <gtest/gtest.h>
+
+#include <rcpputils/time.hpp>
+
+TEST(test_time, test_convert_to_nanoseconds) {
+  rcutils_duration_value_t expect_value = RCUTILS_S_TO_NS(5 * 60);  // 5 minutes
+  rcutils_duration_value_t cast_val;
```
aprotyas, line 21: It looks like OSX CI fails because `cast_val` (may) be uninitialized at `EXPECT_EQ`. The `EXPECT_NO_THROW` macro has an if-block inside it, so it's not guaranteed that `cast_val` will be initialized before reading.

Maybe this instead?
```suggestion
  rcutils_duration_value_t cast_val = 0;
```

aprotyas (changes requested):
Suggesting some inline fixes that should help address the problems with CI.

test/test_time.cpp
Barry-Xu-2018: Agree. thanks

Barry-Xu-2018:
@fujitatomoya   

Please help to re-run CI.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15270/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9986/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12951/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15463/)

aprotyas (approved):

include/rcpputils/time.hpp
```diff
@@ -0,0 +1,62 @@
...
+  // https://stackoverflow.com/a/44637334/2089061
+  // However, this doesn't solve the issue for all possible duration types of period.
+  // Follow-up issue: https://github.com/ros2/rclcpp/issues/1177
+  constexpr auto ns_max_as_double =
+    std::chrono::duration_cast<std::chrono::duration<double, std::chrono::nanoseconds::period>>(
+    maximum_safe_cast_ns);
+  if (time > ns_max_as_double) {
+    throw std::invalid_argument{
+            "time must be less than std::chrono::nanoseconds::max()"};
+  }
```
ivanpauno, line 55: shouldn't we have a check for the minimum? like in the SO answer

clalancette:
This looks like it needs a rebase, then CI and I think we can get it in.

include/rcpputils/time.hpp
```diff
@@ -0,0 +1,71 @@
...
+  // Casting to a double representation might lose precision and allow the check below to succeed
+  // but the actual cast to nanoseconds fail. Using 1 DurationT worth of nanoseconds less than max
+  constexpr auto maximum_safe_cast_ns =
+    std::chrono::nanoseconds::max() - std::chrono::duration<DurationRepT, DurationT>(1);
+  // If period is greater than nanoseconds::max(), the duration_cast to nanoseconds will overflow
+  // a signed integer, which is undefined behavior. Checking whether any std::chrono::duration is
+  // greater than nanoseconds::max() is a difficult general problem. This is a more conservative
+  // version of Howard Hinnant's (the <chrono> guy>) response here:
+  // https://stackoverflow.com/a/44637334/2089061
+  // However, this doesn't solve the issue for all possible duration types of period.
```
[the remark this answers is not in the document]
ivanpauno: I mean, isn't that case detected in the method an a `std::invalid_argument` exception thrown?
If we always throw `std::invalid_argument` when the conversion is not possible we should comment that instead.
If there's UB under some conversions, we should clarify which cases are UB.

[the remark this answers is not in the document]
Barry-Xu-2018: Do you means we should show available range in comments ?  
e.g.  

If input time > 2562047788015 hours (nanoseconds::max) or input time < -2562047788015 hours (nanoseconds::min), conversion fails and exception std::invalid_argument is thrown.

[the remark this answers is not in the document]
Barry-Xu-2018: If input time > 2562047 hours (nanoseconds::max) or input time < -2562047 hours (nanoseconds::min), conversion fails and exception std::invalid_argument is thrown.

include/rcpputils/time.hpp
```diff
@@ -0,0 +1,72 @@
...
+ * std::chrono::nanoseconds::min().
+ */
+template<typename DurationRepT = int64_t, typename DurationT = std::milli>
+std::chrono::nanoseconds convert_to_nanoseconds(
+  const std::chrono::duration<DurationRepT, DurationT> & time)
+{
+  // Casting to a double representation might lose precision and allow the check below to succeed
+  // but the actual cast to nanoseconds fail. Using 1 DurationT worth of nanoseconds less than max
+  constexpr auto maximum_safe_cast_ns =
+    std::chrono::nanoseconds::max() - std::chrono::duration<DurationRepT, DurationT>(1);
```
Barry-Xu-2018, line 42: Found a bug.   
`std::chrono::duration<DurationRepT, DurationT>(1)` depend on user input. This leads `maximum_safe_cast_ns` isn't fixed value.  I will use std::chrono::nanoseconds(1) instead.

iuhilnehc-ynos (approved):
LGTM

iuhilnehc-ynos:
> LGTM

Sorry, wrong place to reply.

include/rcpputils/time.hpp
fujitatomoya: I thought this was not a bug, it is intentionally based on user template unit. if i am not mistaken...

Barry-Xu-2018: If depend on user template unit, there is a below problem.  
For the same time, std::chrono::minutes(2562047*60) doesn't throw exception, but std::chrono::hours(2562047) can show exception.

include/rcpputils/time.hpp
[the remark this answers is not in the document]
ivanpauno: > Do you means we should show available range in comments ?

No, that's already clear [here](https://github.com/ros2/rcpputils/blob/c6acdfaaa8db15f5a0adf0878ccae6a45dd83f77/include/rcpputils/time.hpp#L32-L33).

What I don't understand is the comment "However, this doesn't solve the issue for all possible duration types of period.". 
What duration types doesn't this solve the issue for? What is the "issue"?
If there's a duration type where the checks we're doing aren't going to guarantee that the conversion will either succeed or throw `std::invalid_argument`, we should be specifying for what duration types that can happen.

If the function is always going to either succeed or throw, and there's no posibility of UB for all possible duration types, then we should delete that comment because it's not really clear what it means.

include/rcpputils/time.hpp
```diff
@@ -0,0 +1,72 @@
...
+/**
+ * This function help to convert from std::chrono::duration to std::chrono::nanoseconds and throw
+ * exception if overflow occurs while coverting.
+ *
+ * \param[in] time The time to be converted to std::chrono::nanoseconds.
+ * \return std::chrono::nanoseconds.
+ * \throws std::invalid_argument if time is bigger than std::chrono::nanoseconds::max() or less than
+ * std::chrono::nanoseconds::min().
+ */
+template<typename DurationRepT = int64_t, typename DurationT = std::milli>
```
ivanpauno, line 35: there's no need of defaults, both parameters will be infered based on the argument passed to the function

```suggestion
template<typename DurationRepT, typename DurationT>
```

include/rcpputils/time.hpp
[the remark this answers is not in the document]
Barry-Xu-2018: Thank you for the detailed explanation.    
Yes. This function return success or throw. There are  no other cases.    
I will update comments.

include/rcpputils/time.hpp
```diff
@@ -0,0 +1,72 @@
...
+std::chrono::nanoseconds convert_to_nanoseconds(
+  const std::chrono::duration<DurationRepT, DurationT> & time)
+{
+  // Casting to a double representation might lose precision and allow the check below to succeed
+  // but the actual cast to nanoseconds fail. Using 1 worth of nanoseconds less than max
+  constexpr auto maximum_safe_cast_ns =
+    std::chrono::nanoseconds::max() - std::chrono::nanoseconds(1);
```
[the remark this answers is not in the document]
Barry-Xu-2018: After considering different cases, not find it is necessary.   
I will remove it.

Barry-Xu-2018:
Do rebase

ivanpauno (approved):
lgtm!

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15620/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10301/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13288/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15837/)

PR #158, opened 2022-01-07
Fix include order for cpplint

jacobperron:
Relates to https://github.com/ament/ament_lint/pull/324

### DIFF
--- a/test/test_accumulator.cpp
+++ b/test/test_accumulator.cpp
@@ -20,12 +20,12 @@
 #endif
 #endif
 
-#include <rcppmath/rolling_mean_accumulator.hpp>
-
-#include <gtest/gtest.h>
 #include <cmath>
 #include <memory>
 
+#include "gtest/gtest.h"
+
+#include <rcppmath/rolling_mean_accumulator.hpp>
 
 TEST(TestAccumulator, test_accumulator)
 {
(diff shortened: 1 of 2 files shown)


fujitatomoya (approved):

test/test_accumulator.cpp
```diff
@@ -20,12 +20,12 @@
...
 
-#include <rcppmath/rolling_mean_accumulator.hpp>
-
-#include <gtest/gtest.h>
 #include <cmath>
 #include <memory>
 
+#include "gtest/gtest.h"
+
+#include <rcppmath/rolling_mean_accumulator.hpp>
```
clalancette, line 28: Shouldn't this use the double-quotes instead of the brackets?

test/test_env.cpp
```diff
@@ -12,13 +12,12 @@
...
-#include <gtest/gtest.h>
-
-#include <rcpputils/env.hpp>
-
 #include <stdexcept>
 #include <string>
 
+#include "gtest/gtest.h"
+
+#include <rcpputils/env.hpp>
```
clalancette, line 20: Same question here.

test/test_accumulator.cpp
jacobperron: Now using double-quotes (e2bd343)

test/test_env.cpp
jacobperron: e2bd343

clalancette (approved):

PR #159, opened 2022-01-15
2.3.2

audrow:
I wasn't able to push to master without approval in this repo.


audrow:
Closing as it will be easier to just get write access.

nuclearsandwich (approved):
To make sure that this lands on master as a solitary commit you'll want to use the Squash and Merge or Rebase and Merge options, for a single commit they're basically the same. When you do I'd suggest editing the commit message so that it contains only the version string as is convention for release commits. (See https://github.com/ros2/rcpputils/commit/dfa549ac1ea517da863c996b5d4753a8d0952da9)

PR #163, opened 2022-04-30
2.5.1

audrow:
This PR bumps the minor version to separate Rolling and Humble.

I also bumped the patch version because I accidentally pushed the tag before this PR was merged.

~~(Note, I accidentally pushed the tag 2.5.0 already. I updated my script, but forgot to source my rc file. Hopefully won't happen in the future.)~~

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,9 @@
 Changelog for package rcpputils
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+2.5.1 (2022-05-04)
+------------------
+
 2.5.0 (2022-04-29)
 ------------------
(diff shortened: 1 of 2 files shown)


clalancette:
> (Note, I accidentally pushed the tag 2.5.0 already. I updated my script, but forgot to source my rc file. Hopefully won't happen in the future.)

Unfortunately, that means that we are going to have to do 2.5.1 now.  That's because the tag isn't going to be where we expect it (i.e. on the master branch), but will instead be off on some commit that isn't accessible from any branch.  And we don't want to delete tags that are already pushed.  So I'll suggest closing this out, and doing 2.5.1 instead.

clalancette (approved):

PR #165, opened 2022-08-29
require C++17 and deprecate the rcppmath namespace

wjwwood:
As the title says, now that we can use c++17 in rolling we can do away with the `clamp.hpp` header because it's just implementing `std::clamp` which was added in c++17. Also as discussed in https://github.com/ros2/rcpputils/pull/85#issuecomment-754134335, removing the `rcppmath` namespace as it doesn't follow our conventions that packages should only put headers and symbols in their own namespace unless at all possible.

### DIFF
--- a/include/rcppmath/clamp.hpp
+++ b/include/rcppmath/clamp.hpp
@@ -21,11 +21,6 @@
 
 #include <cassert>
 
-
-#ifndef RCPPMATH__CLAMP_HPP_WARNING_DISABLED
-# warning "the rcppmath namespace is deprecated, include <algorithm> and use std::clamp instead"
-#endif
-
 namespace rcppmath
 {
 /**
@@ -42,6 +37,7 @@ namespace rcppmath
  *  a dangling reference if that parameter is returned.
  */
 template<class T>
+[[deprecated("use std::clamp instead, introduced in C++17")]]
 constexpr const T & clamp(const T & v, const T & lo, const T & hi)
 {
   assert(!(hi < lo) );
@@ -63,6 +59,7 @@ constexpr const T & clamp(const T & v, const T & lo, const T & hi)
  * \sa rcppmath::clamp(const T&, const T&, const T&)
  */
 template<class T, class Compare>
+[[deprecated("use std::clamp instead, introduced in C++17")]]
 constexpr const T & clamp(const T & v, const T & lo, const T & hi, Compare comp)
 {
   assert(!comp(hi, lo) );
--- a/include/rcppmath/clamp.hpp
+++ b/include/rcppmath/clamp.hpp
@@ -21,6 +21,11 @@
 
 #include <cassert>
 
+
+#ifndef RCPPMATH__CLAMP_HPP_WARNING_DISABLED
+# warning "the rcppmath namespace is deprecated, include <algorithm> and use std::clamp instead"
+#endif
+
 namespace rcppmath
 {
 /**
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,9 +10,9 @@ find_package(rcutils REQUIRED)
 if(NOT CMAKE_C_STANDARD)
   set(CMAKE_C_STANDARD 11)
 endif()
-# Default to C++14
+# Default to C++17
 if(NOT CMAKE_CXX_STANDARD)
-  set(CMAKE_CXX_STANDARD 14)
+  set(CMAKE_CXX_STANDARD 17)
 endif()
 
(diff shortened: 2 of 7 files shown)


include/rcppmath/clamp.hpp
```diff
@@ -12,57 +12,17 @@
...
- * a dangling reference if that parameter is returned.
- * \sa rcppmath::clamp(const T&, const T&, const T&)
- */
-template<class T, class Compare>
-constexpr const T & clamp(const T & v, const T & lo, const T & hi, Compare comp)
-{
-  assert(!comp(hi, lo) );
-  return comp(v, lo) ? lo : comp(hi, v) ? hi : v;
-}
+using rcpputils::clamp;
```
wjwwood, line 25: Note that it wasn't possible (as far as I could figure out) to deprecate this using statement, but the actual clamp functions this includes here are now deprecated in favor of `std::clamp` so it I didn't try to figure it our further.

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17190/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11731/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17612/)

wjwwood:
@ros-pull-request-builder please test this

wjwwood:
@ros-pull-request-builder test this please

nuclearsandwich:
I've marked the agent that was exhibiting this issue offline to test.

@ros-pull-request-builder retest this please.

cottsay:
@ros-pull-request-builder retest this please.

nuclearsandwich:
> @ros-pull-request-builder retest this please.

With this build the result is now UNSTABLE rather than failed and the infrastructure issues should be resolved.

include/rcpputils/clamp.hpp
```diff
@@ -0,0 +1,75 @@
...
+ * \param[in] lo The lower boundary.
+ * \param[in] hi The higher boundary.
+ * \return Reference to lo if v is less than lo, reference to hi if hi is less than v, otherwise
+ * reference to v.
+ * \note Implementation from https://en.cppreference.com/w/cpp/algorithm/clamp.
+ * \warning Capturing the result of clamp by reference if one of the parameters is rvalue produces
+ *  a dangling reference if that parameter is returned.
+ */
+template<class T>
+[[deprecated("use std::clamp instead, introduced in C++17")]]
```
clalancette, line 42: It's kind of odd to be adding a function that is immediately deprecated.

If the goal here is to completely rely on `std::clamp` instead of this version, then I'll suggest that we do *not* move this header out of `rcppmath` for now, but instead just mark it deprecated in there.  Any downstream users will have to change from `rcppmath::clamp` to `std::clamp` directly, and then in J-Turtle we can just remove everything under the `rcppmath` namespace.

clalancette (changes requested):

include/rcpputils/clamp.hpp
wjwwood: But does it hurt to move it and deprecate it in favor of `std::clamp`? I would just like to completely gut that namespace. I could also see that we "tock" these two things separately. I.e. we could drop the `rcppmath` namespace in the next release, but for some reason not move forward with the use of `std::clamp`.

If you really don't want it as-is, I would prefer to just move the namespace and leave the `std::clamp` thing for later. I only did that here because I noticed as I was moving the functions. The primary goal here is to deprecate the `rcppmath` namespace.

wjwwood: Ok, I still don't really see how undoing changes in this pr makes it better, but I went a head and did it. I had to rebase and stuff, so everything is in [7ec674f](https://github.com/ros2/rcpputils/pull/165/commits/7ec674fec97965128ee0e79e6b32b7486affdcf4) now, including `uncrustify` fixes.

wjwwood: Just to recap, I think I've updated it as you suggested. I think that it's fine in this state, so if you agree we can merge it as-is (with CI)?

wjwwood:
New CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17484/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12011/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18032/)

clalancette (approved):
Looks good with green CI.

wjwwood:
The CI test failure on Windows is unrelated, merging.

PR #180, opened 2023-08-30
Add unique_lock implementation with clang thread safety annotations

emersonknapp:
Can be used to replace rosbag2 implementation https://github.com/ros2/rosbag2/blob/rolling/rosbag2_cpp/src/rosbag2_cpp/clocks/time_controller_clock.cpp#L31
and false positive errors in RMW builds like https://ci.ros2.org/view/nightly/job/nightly_linux_clang_libcxx/1628/clang/new/source.cf52365c-c9dc-4c3e-8639-55ffa17e1b64/#115

### DIFF
--- a/include/rcpputils/unique_lock.hpp
+++ b/include/rcpputils/unique_lock.hpp
@@ -1,0 +1,36 @@
+// Copyright 2023 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "rcpputils/thread_safety_annotations.hpp"
+
+namespace rcpputils
+{
+
+/**
+ * @brief Trivial std::unique_lock wrapper providing constructor that allows Clang's
+ * Thread Safety Analysis.
+ * The libc++ std::unique_lock does not have these annotations.
+ */
+template<typename MutexT>
+class RCPPUTILS_TSA_SCOPED_CAPABILITY unique_lock : public std::unique_lock<MutexT>
+{
+public:
+  explicit unique_lock(std::mutex & mu) RCPPUTILS_TSA_ACQUIRE(mu)
+  : std::unique_lock<MutexT>(mu)
+  {}
+
+  ~TSAUniqueLock() RCPPUTILS_TSA_RELEASE() {}
+};
+
+}  // namespace rcpputils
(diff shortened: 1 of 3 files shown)


emersonknapp:
Probably needs a test

fujitatomoya (approved):
having this in rcpputil makes sense, test would be ideal.

clalancette:
And just for verification, it looks like using this instead of `std::unique_lock` in `rmw_fastrtps` makes a difference: https://github.com/ros2/rmw_fastrtps/pull/712#issuecomment-1701294278 (there are about 140 fewer violations than on e.g. https://ci.ros2.org/view/nightly/job/nightly_linux_clang_libcxx/1629/)

include/rcpputils/unique_lock.hpp
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * @brief Trivial std::unique_lock wrapper providing constructor that allows Clang's
+ * Thread Safety Analysis.
+ * The libc++ std::unique_lock does not have these annotations.
+ */
+template<typename MutexT>
+class RCPPUTILS_TSA_SCOPED_CAPABILITY unique_lock : public std::unique_lock<MutexT>
+{
+public:
+  explicit unique_lock(std::mutex & mu) RCPPUTILS_TSA_ACQUIRE(mu)
```
clalancette, line 34: I'm not 100% sure of this, but shouldn't this be:

```suggestion
  explicit unique_lock(MutexT & mu) RCPPUTILS_TSA_ACQUIRE(mu)
```

clalancette (commented):
I've left a possible bug inline.

As far as a test goes, I'm actually not entirely sure what we would test here.  More-or-less it would be testing whether `std::unique_lock` works.  Any thoughts from @emersonknapp or @fujitatomoya on what kind of test you think we should do?

include/rcpputils/unique_lock.hpp
emersonknapp: Yep, I need to add a test with two kinds of mutex before we merge this!

fujitatomoya:
@clalancette that is actually good question. i was thinking that we can simulate https://ci.ros2.org/view/nightly/job/nightly_linux_clang_libcxx/1628/clang/new/source.cf52365c-c9dc-4c3e-8639-55ffa17e1b64/#115 but it actually is the test for `RCPPUTILS_TSA_SCOPED_CAPABILITY`.

we could test `rcpputils::unique_lock` works, but i dont think that is even necessary either.

clalancette:
@emersonknapp When you get a chance, can you rebase this to fix the DCO bot?

emersonknapp:
Added a simple compile-test that we can use multiple types of mutex.

emersonknapp:
Pulls: ros2/rcpputils#180
Gist: https://gist.githubusercontent.com/emersonknapp/c4a008cb46d56f9548cffbcbaefac5a1/raw/22101fab9d965b83665b049428add3a1a64e682a/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/12608
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19455/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13978/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20178/)

clalancette (approved):

PR #196, opened 2024-07-08
Deprecated path class

ahcorde:
We are using C++17, so we can deprecate `path` class.

This might require some changes downstream


include/rcpputils/filesystem_helper.hpp
```diff
@@ -352,6 +360,58 @@ RCPPUTILS_PUBLIC bool operator!=(const path & a, const path & b);
...
+// RCPPUTILS_PUBLIC bool operator!=(const path & a, const path & b);
+
+/**
+* \brief Convert the path to a string for ostream usage, such as in logging or string formatting.
+*
+* \param[in] os The stream to send the path string to
+* \param[in] p The path to stringify
+* \return The ostream, for chaining
+*/
+RCPPUTILS_PUBLIC std::ostream & operator<<(std::ostream & os, const std::filesystem::path & p);
```
clalancette, line 413: I'm a bit confused about this entire API.  Why can't we just use the existing` rcpputils::fs::create_temp_directory`?

ahcorde: There is an ambiguous method call

clalancette: > There is an ambiguous method call

Can you explain more about it?  If we use the fully-qualified version, I'd think that `rcpputils::fs::create_temp_directory` should be sufficient, no?

ahcorde: `rcpputils::fs::create_temp_directory` return a rcpputils::fs::path and the idea is to return a `std::filesystem::path`.

 Then the first argument is a string and we got an ambiguous error, that's why I need to change the signature of the function.

from `create_temp_directory` to `create_temporal_directory`

clalancette: > `rcpputils::fs::create_temp_directory` return a rcpputils::fs::path and the idea is to return a `std::filesystem::path`.

Ah, I see!  I missed that fact, thanks.  Now I see why we need another function name.  All right, I have some other comments here, but I'll leave those separately.

include/rcpputils/filesystem_helper.hpp
```diff
@@ -352,6 +360,58 @@ RCPPUTILS_PUBLIC bool operator!=(const path & a, const path & b);
...
+/**
+ * \brief Get a path to a location in the temporary directory, if it's available.
+ *
+ * This does not create any directories.
+ * On Windows, this uses "GetTempPathA"
+ * On non-Windows, this prefers the environment variable TMPDIR, falling back to /tmp
+ *
+ * \return A path to a directory for storing temporary files and directories.
+ */
+RCPPUTILS_PUBLIC std::filesystem::path temporal_directory_path();
```
clalancette, line 379: While "temporal" isn't exactly wrong, it is also not a common name for this.  The word "temporary" is much more common.  So I'm going to suggest that we change it:

```suggestion
RCPPUTILS_PUBLIC std::filesystem::path temporary_directory_path();
```

include/rcpputils/filesystem_helper.hpp
```diff
@@ -352,6 +360,58 @@ RCPPUTILS_PUBLIC bool operator!=(const path & a, const path & b);
...
+ *
+ * \param[in] base_name User-specified portion of the created directory
+ * \param[in] parent_path The parent path of the directory that will be created
+ * \return A path to a newly-created directory with base_name and a 6-character unique suffix
+ *
+ * \throws std::system_error If any OS APIs do not succeed.
+ */
+RCPPUTILS_PUBLIC std::filesystem::path create_temporal_directory(
+  const std::string & base_name,
+  const std::filesystem::path & parent_path = temporal_directory_path());
```
clalancette, line 396:
```suggestion
RCPPUTILS_PUBLIC std::filesystem::path create_temporary_directory(
  const std::string & base_name,
  const std::filesystem::path & parent_path = temporary_directory_path());
```

include/rcpputils/filesystem_helper.hpp
```diff
@@ -352,6 +360,58 @@ RCPPUTILS_PUBLIC bool operator!=(const path & a, const path & b);
...
+
+// /**
+//  * \brief Compare two paths for equality.
+//  *
+//  * \return True if both paths are equal as strings.
+//  */
+// RCPPUTILS_PUBLIC bool operator==(const path & a, const path & b);
+// RCPPUTILS_PUBLIC bool operator!=(const path & a, const path & b);
```
clalancette, line 404:
```suggestion
```

include/rcpputils/filesystem_helper.hpp
clalancette, line 413: Is this stream operator needed/implemented?

clalancette (changes requested):

include/rcpputils/filesystem_helper.hpp
```diff
@@ -65,14 +66,21 @@ static constexpr const char kPreferredSeparator = RCPPUTILS_IMPL_OS_DIRSEP;
...
 
-
+// TODO(ahcorde): Remove deprecated class on the next release.
+#if !defined(_WIN32)
+# pragma GCC diagnostic push
+# pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+#else  // !defined(_WIN32)
+# pragma warning(push)
+# pragma warning(disable: 4996)
+#endif
```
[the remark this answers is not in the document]
ahcorde: I put this block here because not only the class is generating the deprecation warnings,  there are some  functions also like `remove_extension`, `create_directories`, etc.

include/rcpputils/filesystem_helper.hpp
ahcorde: removed

include/rcpputils/filesystem_helper.hpp
```diff
@@ -352,6 +360,40 @@ RCPPUTILS_PUBLIC bool operator!=(const path & a, const path & b);
...
+ *
+ * \param[in] base_name User-specified portion of the created directory
+ * \param[in] parent_path The parent path of the directory that will be created
+ * \return A path to a newly-created directory with base_name and a 6-character unique suffix
+ *
+ * \throws std::system_error If any OS APIs do not succeed.
+ */
+RCPPUTILS_PUBLIC std::filesystem::path create_temporary_directory(
+  const std::string & base_name,
+  const std::filesystem::path & parent_path = temporary_directory_path());
```
clalancette, line 396: I poked at this a while today with @mjcarroll , and tried to come up with a nicer way to do this.  We really couldn't figure out a better way than making a new function name.

However, I pulled down all of the source code released into Rolling.  And I found that `rcpputils::fs::temp_directory_path` and `rcpputils::fs::create_temp_directory` aren't used anywhere outside of the core.  In the core, they are only used in:

* `rcpputils` - the implementation and the tests
* `rosbag2_compression` - some tests
* `rosbag2_cpp` - some tests
* `rosbag2_storage_mcap` - a test
* `rosbag2_storage_sqlite3` - a test
* `rosbag2_transport` - a test
* `rosbag2_test_common` - a test
* `rcl_logging_spdlog` - a test

Thus, at the cost of only 3 PRs (this one, one in `rcl_logging`, and one in `rosbag2`), we can completely break this API.  I think we should do that.  In particular, I think we should remove `temporary_directory_path` and `create_temporary_directory`, and also remove `temp_directory_path`.  Then we change `create_temp_directory` to take in a `std::filesystem::path`, and also return a `std::filesystem::path`.  And then we update the tests here to use `std::filesystem::path`, as well as the tests in `rcl_logging` and `rosbag2`.  At the end of those 3 PRs (which will have to be merged at the same time), we'll have the API we want going forward.

Does that make sense to you @ahcorde?

clalancette: Actually, in point of fact, that is a bit enough change that I think we should do that on its own, separate from this PR.  Once that is in, we can revisit deprecating `rcpputils::fs::path` separately (i.e. this PR).

ahcorde: Related PR https://github.com/ros2/rcpputils/pull/197

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21521/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15803/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1019/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22237/)

Included https://github.com/ros2/rclcpp/pull/2579

test/test_filesystem_helper_std.cpp
```diff
@@ -0,0 +1,520 @@
...
+
+#ifdef _WIN32
+static constexpr const bool is_win32 = true;
+#else
+static constexpr const bool is_win32 = false;
+#endif
+
+using path = std::filesystem::path;
+
+std::string build_extension_path()
```
clalancette, line 31:
```suggestion
static const std::string build_extension_path()
```

test/test_filesystem_helper_std.cpp
```diff
@@ -0,0 +1,520 @@
...
+{
+  return is_win32 ? R"(C:\foo\bar\baz.yml)" : "/bar/foo/baz.yml";
+}
+
+std::string build_double_extension_path()
```
clalancette, line 36:
```suggestion
static const std::string build_double_extension_path()
```

test/test_filesystem_helper_std.cpp
```diff
@@ -0,0 +1,520 @@
...
+{
+  return is_win32 ? R"(C:\bar\baz.bar.yml)" : "/foo/baz.bar.yml";
+}
+
+std::string build_no_extension_path()
```
clalancette, line 41:
```suggestion
static const std::string build_no_extension_path()
```

test/test_filesystem_helper_std.cpp
```diff
@@ -0,0 +1,520 @@
...
+{
+  return is_win32 ? R"(.\test_folder)" : R"(./test_folder)";
+}
+
+std::string build_directory_path()
```
clalancette, line 46:
```suggestion
static const std::string build_directory_path()
```

test/test_filesystem_helper_std.cpp
```diff
@@ -0,0 +1,520 @@
...
+      }
+      {
+        auto p = path("\\");
+        EXPECT_EQ(p.parent_path().string(), "\\");
+      }
+    } else {
+      auto p = path("/");
+      EXPECT_EQ(p.parent_path().string(), "/");
+    }
+  }
```
clalancette, line 119: The extra scope here doesn't seem necessary, and I think we can combine these into a shorter set of `if` statements:

```suggestion
  if (is_win32) {
    {
      auto p = path("C:\\foo");
      EXPECT_EQ(p.parent_path().string(), "C:\\");
    }
    {
      auto p = path("\\foo");
      EXPECT_EQ(p.parent_path().string(), "\\");
    }
    {
      auto p = path("C:\\");
      EXPECT_EQ(p.parent_path().string(), "C:\\");
    }
    {
      auto p = path("\\");
      EXPECT_EQ(p.parent_path().string(), "\\");
    }
  } else {
    {
      auto p = path("/foo");
      EXPECT_EQ(p.parent_path().string(), "/");
    }
    {
      auto p = path("/");
      EXPECT_EQ(p.parent_path().string(), "/");
    }
  }
```

test/test_filesystem_helper_std.cpp
```diff
@@ -0,0 +1,520 @@
...
+/**
+ * Test filesystem manipulation API.
+ *
+ * NOTE: expects the current directory to be write-only, else test will fail.
+ *
+ */
+TEST(TestFilesystemHelper, filesystem_manipulation)
+{
+  auto dir = path(build_directory_path());
+  (void)std::filesystem::remove_all(dir);
```
clalancette, line 283:
```suggestion
  std::filesystem::remove_all(dir);
```

test/test_filesystem_helper_std.cpp
```diff
@@ -0,0 +1,520 @@
...
+    expected_file_size = static_cast<uint64_t>(output_buffer.tellp());
+  }
+
+  EXPECT_TRUE(std::filesystem::exists(file));
+  EXPECT_TRUE(std::filesystem::is_regular_file(file));
+  EXPECT_FALSE(std::filesystem::is_directory(file));
+  std::error_code ec;
+  EXPECT_FALSE(std::filesystem::create_directories(file, ec));
+  EXPECT_GE(std::filesystem::file_size(file), expected_file_size);
+  EXPECT_THROW((void)std::filesystem::file_size(dir), std::filesystem::filesystem_error) <<
```
clalancette, line 304:
```suggestion
  EXPECT_THROW(std::filesystem::file_size(dir), std::filesystem::filesystem_error) <<
```

test/test_filesystem_helper_std.cpp
```diff
@@ -0,0 +1,520 @@
...
+    "file_size is only applicable for files!";
+  EXPECT_FALSE(std::filesystem::remove(dir, ec));
+  EXPECT_TRUE(std::filesystem::remove(file, ec));
+  EXPECT_THROW((void)std::filesystem::file_size(file), std::filesystem::filesystem_error);
```
clalancette, line 308:
```suggestion
  EXPECT_THROW(std::filesystem::file_size(file), std::filesystem::filesystem_error);
```

clalancette (changes requested):

test/test_filesystem_helper_std.cpp
ahcorde: removing this, it will generate a warning because we are not using the result of the call

clalancette: Oh, right.  We have to leave this one (and the other `file_size` one) in place, sorry about that.

test/test_filesystem_helper_std.cpp
ahcorde: https://github.com/ros2/rcpputils/pull/196/commits/5467b2afee010a87b82aa3701c88bf2a31efcce0

test/test_filesystem_helper_std.cpp
ahcorde: https://github.com/ros2/rcpputils/pull/196/commits/652764cca26071b11fbbc2fbf8c1a35324033d6e

test/test_filesystem_helper_std.cpp
ahcorde: https://github.com/ros2/rcpputils/pull/196/commits/652764cca26071b11fbbc2fbf8c1a35324033d6e

test/test_filesystem_helper_std.cpp
ahcorde: https://github.com/ros2/rcpputils/pull/196/commits/652764cca26071b11fbbc2fbf8c1a35324033d6e

test/test_filesystem_helper_std.cpp
ahcorde: https://github.com/ros2/rcpputils/pull/196/commits/652764cca26071b11fbbc2fbf8c1a35324033d6e

test/test_filesystem_helper_std.cpp
ahcorde: https://github.com/ros2/rcpputils/pull/196/commits/652764cca26071b11fbbc2fbf8c1a35324033d6e

test/test_filesystem_helper_std.cpp
```diff
@@ -0,0 +1,519 @@
+// Copyright 2019 Open Source Robotics Foundation, Inc.
```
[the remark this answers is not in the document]
clalancette: Sorry, just realized that @fujitatomoya's comment here is still open.

clalancette (changes requested):
One last thing to change here, then this should be good.

test/test_filesystem_helper_std.cpp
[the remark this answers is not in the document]
ahcorde: https://github.com/ros2/rcpputils/pull/196/commits/a731c8b3c2e36a002510007cda255c33c3c32c1d

clalancette (approved):
Looks good to me with green CI.

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21531/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15811/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1026/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22245/)

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21536/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15820/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1031/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22253/)

PR #197, opened 2024-07-10
Use std::filesystem in create_temp_directory and temp_directory_path

ahcorde:
Related to this comment https://github.com/ros2/rcpputils/pull/196#discussion_r1671225326

update `create_temp_directory` and `temp_directory_path` to use `std::filesystem::path`

Related PRs
 - https://github.com/ros2/rcl_logging/pull/117
 - https://github.com/ros2/rosbag2/pull/1740

### DIFF
--- a/include/rcpputils/filesystem_helper.hpp
+++ b/include/rcpputils/filesystem_helper.hpp
@@ -40,6 +40,7 @@
 #define RCPPUTILS__FILESYSTEM_HELPER_HPP_
 
 #include <cstdint>
+#include <filesystem>
 #include <string>
 #include <vector>
 
@@ -269,7 +270,7 @@ RCPPUTILS_PUBLIC bool exists(const path & path_to_check);
  *
  * \return A path to a directory for storing temporary files and directories.
  */
-RCPPUTILS_PUBLIC path temp_directory_path();
+RCPPUTILS_PUBLIC std::filesystem::path temp_directory_path();
 
 /**
  * \brief Construct a uniquely named temporary directory, in "parent", with format base_nameXXXXXX
@@ -284,9 +285,9 @@ RCPPUTILS_PUBLIC path temp_directory_path();
  *
  * \throws std::system_error If any OS APIs do not succeed.
  */
-RCPPUTILS_PUBLIC path create_temp_directory(
+RCPPUTILS_PUBLIC std::filesystem::path create_temp_directory(
   const std::string & base_name,
-  const path & parent_path = temp_directory_path());
+  const std::filesystem::path & parent_path = temp_directory_path());
 
 /**
  * \brief Return current working directory.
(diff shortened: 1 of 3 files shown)


include/rcpputils/filesystem_helper.hpp
```diff
@@ -269,7 +270,7 @@ RCPPUTILS_PUBLIC bool exists(const path & path_to_check);
  *
  * \return A path to a directory for storing temporary files and directories.
  */
-RCPPUTILS_PUBLIC path temp_directory_path();
+RCPPUTILS_PUBLIC std::filesystem::path temp_directory_path();
```
clalancette, line 273: I'm going to suggest that we just remove this API completely, as it already exists in [std::filesystem](https://en.cppreference.com/w/cpp/filesystem/temp_directory_path):

```suggestion
```

src/filesystem_helper.cpp
```diff
@@ -284,7 +284,7 @@ bool exists(const path & path_to_check)
   return path_to_check.exists();
 }
 
-path temp_directory_path()
+std::filesystem::path temp_directory_path()
```
clalancette, line 287: Same here; just remove this implementation.

clalancette (changes requested):

include/rcpputils/filesystem_helper.hpp
```diff
@@ -259,35 +260,6 @@ RCPPUTILS_PUBLIC uint64_t file_size(const path & p);
...
- * \param[in] parent_path The parent path of the directory that will be created
- * \return A path to a newly-created directory with base_name and a 6-character unique suffix
- *
- * \throws std::system_error If any OS APIs do not succeed.
- */
-RCPPUTILS_PUBLIC path create_temp_directory(
-  const std::string & base_name,
-  const path & parent_path = temp_directory_path());
-
 /**
```
clalancette, line 263: So we still need to keep `create_temp_directory`, since there are downstream users.  We want to change that to use `std::filesystem::path`.

But we don't need `temp_directory_path`, since that is provided by `std::filesystem` already.

clalancette (changes requested):

include/rcpputils/filesystem_helper.hpp
ahcorde: yeah, sorry I was to excited about removing code. Restored

clalancette (commented):
This implementation generally looks good to me.  I'm going to hold off on approving it until we figure out what is going on with https://github.com/ros2/rosbag2/pull/1740

MichaelOrlov (commented):
The implementation has API-breaking changes and still uses platform-dependent code, which is not good.
Creating a new function with a different name and deprecating the old one would be better.

mjcarroll:
> still uses platform-dependent code

I don't believe there is a way around this.  Using posix `mkdtmp` is probably as good as we are going to get here.

Should we perhaps throw or return error code on non-implemented platforms?

ahcorde:
As discussed in the waffle meeting, we will close this issue and @MichaelOrlov will create a new PR.

PR #198, opened 2024-07-11
Replace create_temp_directory with the new create_temporary_directory

MichaelOrlov:
- This PR replaces #197 and relates to this comment: https://github.com/ros2/rcpputils/pull/196#discussion_r1671225326
- The newly added `create_temporary_directory(..)` uses std::filesystem::path and doesn't have platform-specific code.
- Also deprecate `create_temp_directory(..)` and `temp_directory_path`

### DIFF
--- a/src/filesystem_helper.cpp
+++ b/src/filesystem_helper.cpp
@@ -46,6 +46,7 @@
 #include <random>
 #include <string>
 #include <system_error>
+#include <stdexcept>
 #include <vector>
 
 #ifdef _WIN32
@@ -345,6 +346,11 @@ path create_temp_directory(const std::string & base_name, const path & parent_pa
 std::filesystem::path create_temporary_directory(
   const std::string & base_name, const std::filesystem::path & parent_path, size_t max_tries)
 {
+  auto it =
+    std::find(base_name.begin(), base_name.end(), std::filesystem::path::preferred_separator);
+  if (it != base_name.end()) {
+    throw std::invalid_argument("The base_name contain directory-separator");
+  }
   // mersenne twister random generator engine seeded with the std::random_device
   std::mt19937 random_generator(std::random_device{}());
   std::uniform_int_distribution<> distribution(0, 999999);
--- a/src/filesystem_helper.cpp
+++ b/src/filesystem_helper.cpp
@@ -346,14 +346,12 @@ path create_temp_directory(const std::string & base_name, const path & parent_pa
 std::filesystem::path create_temporary_directory(
   const std::string & base_name, const std::filesystem::path & parent_path, size_t max_tries)
 {
-  auto it =
-    std::find(base_name.begin(), base_name.end(), std::filesystem::path::preferred_separator);
-  if (it != base_name.end()) {
+  if (base_name.find(std::filesystem::path::preferred_separator) != std::string::npos) {
     throw std::invalid_argument("The base_name contain directory-separator");
   }
   // mersenne twister random generator engine seeded with the std::random_device
   std::mt19937 random_generator(std::random_device{}());
-  std::uniform_int_distribution<> distribution(0, 999999);
+  std::uniform_int_distribution<> distribution(0, 0xFFFFFF);
   std::filesystem::path path_to_temp_dir;
   constexpr size_t kSuffixLength = 7;  // 6 chars + 1 null terminator
   char random_suffix_str[kSuffixLength];
--- a/src/filesystem_helper.cpp
+++ b/src/filesystem_helper.cpp
@@ -43,6 +43,7 @@
 #include <algorithm>
 #include <climits>
 #include <cstring>
+#include <random>
 #include <string>
 #include <system_error>
 #include <vector>
@@ -339,6 +340,33 @@ path create_temp_directory(const std::string & base_name, const path & parent_pa
 #endif
 
   return final_path;
+}
+
+std::filesystem::path create_temporary_directory(
+  const std::string & base_name, const std::filesystem::path & parent_path, size_t max_tries)
+{
+  // mersenne twister random generator engine seeded with the std::random_device
+  std::mt19937 random_generator(std::random_device{}());
+  std::uniform_int_distribution<> distribution(0, 999999);
+  std::filesystem::path path_to_temp_dir;
+  constexpr size_t kSuffixLength = 7;  // 6 chars + 1 null terminator
+  char random_suffix_str[kSuffixLength];
+  size_t current_iteration = 0;
+  while (true) {
+    snprintf(random_suffix_str, kSuffixLength, "%06x", distribution(random_generator));
+    const std::string random_dir_name = base_name + random_suffix_str;
+    path_to_temp_dir = parent_path / random_dir_name;
+    // true if the directory was newly created.
+    if (std::filesystem::create_directories(path_to_temp_dir)) {
+      break;
+    }
+    if (current_iteration == max_tries) {
+      throw std::runtime_error(
+        "Exceeded maximum allowed iterations to find non-existing directory");
+    }
+    current_iteration++;
+  }
+  return path_to_temp_dir;
 }
 
 path current_path()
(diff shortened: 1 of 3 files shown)


MichaelOrlov:
cc: @ahcorde, @clalancette - ready for review.

test/test_filesystem_helper.cpp
```diff
@@ -477,16 +480,23 @@ TEST(TestFilesystemHelper, create_temp_directory)
...
   {
     if (is_win32) {
-      EXPECT_THROW(rcpputils::fs::create_temp_directory("illegalchar?"), std::system_error);
+      EXPECT_THROW(rcpputils::fs::create_temporary_directory("illegalchar?"), std::system_error);
     } else {
-      EXPECT_THROW(rcpputils::fs::create_temp_directory("base/name"), std::system_error);
+      std::filesystem::path tmpdir_with_slash_in_base;
+      EXPECT_NO_THROW(
+        tmpdir_with_slash_in_base = rcpputils::fs::create_temporary_directory("base/name"));
+      EXPECT_NO_THROW(std::filesystem::remove_all(tmpdir_with_slash_in_base));
```
MichaelOrlov, line 488: Note: In the new implementation, slashes in the base name is valid case and will be treated as path delimiters when creating folders.

test/test_filesystem_helper.cpp
```diff
@@ -495,15 +505,14 @@ TEST(TestFilesystemHelper, create_temp_directory)
   // edge case inputs
   {
     // Provided no base name we should still get a path with the 6 unique template chars
-    const auto tmpdir_emptybase = rcpputils::fs::create_temp_directory("");
+    const auto tmpdir_emptybase = rcpputils::fs::create_temporary_directory("");
     EXPECT_EQ(tmpdir_emptybase.filename().string().size(), 6u);
 
-    // Empty path doesn't exist and cannot be created
-    EXPECT_THROW(rcpputils::fs::create_temp_directory("basename", path()), std::system_error);
-
```
MichaelOrlov, line 503: Note: The empty path is a valid case for the new implementation. The temp folder will be created in the current folder. 
However, that is not good for tests when creating a new temporary folder from where the tests running. It could be a read-only volume or require special privileges for write access.

test/test_filesystem_helper.cpp
clalancette: I'm not a huge fan of this behavior.  The base name is really the "template" for the name, and as such, it shouldn't be involved in creating additional subdirectories.  That seems like surprising behavior to me, and different than e.g. `mkdtemp` on Linux (which fails if the basename has a `/` in it).  So I'd suggest that if `base_name` contains a directory separator, we throw an exception.

src/filesystem_helper.cpp
```diff
@@ -341,6 +342,33 @@ path create_temp_directory(const std::string & base_name, const path & parent_pa
...
+  while (true) {
+    snprintf(random_suffix_str, kSuffixLength, "%06x", distribution(random_generator));
+    const std::string random_dir_name = base_name + random_suffix_str;
+    path_to_temp_dir = parent_path / random_dir_name;
+    // true if the directory was newly created.
+    if (std::filesystem::create_directories(path_to_temp_dir)) {
+      break;
+    }
+    if (current_iteration == max_tries) {
+      throw std::runtime_error(
```
clalancette, line 364: Nit: we need to `#include <stdexcept>` for `std::runtime_error`.

clalancette (changes requested):
Overall, looks good.  I've left two things inline to fix.

This will also have to be paired with PRs to `rcl_logging` and `rosbag2` so that they use the non-deprecated API.

test/test_filesystem_helper.cpp
MichaelOrlov: Ok, I added a throwing std::invalid_argument exception if base_name contains a directory separator.

src/filesystem_helper.cpp
MichaelOrlov: Added.

MichaelOrlov:
@clalancette, I addressed your comments.

src/filesystem_helper.cpp
```diff
@@ -341,6 +343,38 @@ path create_temp_directory(const std::string & base_name, const path & parent_pa
   return final_path;
 }
 
+std::filesystem::path create_temporary_directory(
```
ahcorde, line 346: #include `<filesystem>`

ahcorde (changes requested):
Just a small nit, otherwise LGTM

src/filesystem_helper.cpp
```diff
@@ -341,6 +343,38 @@ path create_temp_directory(const std::string & base_name, const path & parent_pa
...
+  const std::string & base_name, const std::filesystem::path & parent_path, size_t max_tries)
+{
+  auto it =
+    std::find(base_name.begin(), base_name.end(), std::filesystem::path::preferred_separator);
+  if (it != base_name.end()) {
```
clalancette, line 351: I think we can make this simpler by using the `string.find` method:

```suggestion
  if (base_name.find(std::filesystem::path::preferred_separator) != base_name.npos) {
```

src/filesystem_helper.cpp
```diff
@@ -341,6 +343,38 @@ path create_temp_directory(const std::string & base_name, const path & parent_pa
...
+    throw std::invalid_argument("The base_name contain directory-separator");
+  }
+  // mersenne twister random generator engine seeded with the std::random_device
+  std::mt19937 random_generator(std::random_device{}());
+  std::uniform_int_distribution<> distribution(0, 999999);
```
clalancette, line 356: I think this interval is too small, now that I look at it.  When I ran some tests locally, I always got `0` as the first digit.  And that makes sense, because 999999 == F423F hex, so there will never be something in the first digit.  For 6 digits of hex, we should make this 0xFFFFFF:

```suggestion
  std::uniform_int_distribution<> distribution(0, 0xFFFFFF);
```

clalancette (changes requested):

src/filesystem_helper.cpp
MichaelOrlov: I've added it to the relevant header file. Therefore no need to add it to the cpp file.

src/filesystem_helper.cpp
MichaelOrlov: Fixed in the https://github.com/ros2/rcpputils/pull/198/commits/8b4087d95df458948f20f0d533c56330d5733606 commit

src/filesystem_helper.cpp
MichaelOrlov: @clalancette Good catch! I forgot that we convert in HEX and that in hex the range should be bigger.
Fixed in the https://github.com/ros2/rcpputils/pull/198/commits/8b4087d95df458948f20f0d533c56330d5733606 commit.

ahcorde (approved):

MichaelOrlov:
Pulls: ros2/rcpputils#198
Gist: https://gist.githubusercontent.com/MichaelOrlov/053726535ccac0098e0f98cd4d8aa796/raw/b89be092beb15684cb63ffbb1ffd8f9203be2c36/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14241
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21484/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15768/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/983/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22199/)

MichaelOrlov:
Re-run Windows CI after addressing build failures
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22213/)

ahcorde:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21517/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15799/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1013/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22233/)

Included https://github.com/ros2/rosbag2/pull/1740/ and https://github.com/ros2/rcl_logging/pull/117

PR #201, opened 2024-11-14
fix memory leak for remove_all().

fujitatomoya:
closes https://github.com/ros2/rcpputils/issues/200

### DIFF
--- a/src/filesystem_helper.cpp
+++ b/src/filesystem_helper.cpp
@@ -64,6 +64,7 @@
 #endif
 
 #include "rcutils/env.h"
+#include "rcpputils/scope_exit.hpp"
 #include "rcpputils/split.hpp"
 
 namespace rcpputils
@@ -474,6 +475,10 @@ bool remove_all(const path & p)
   return 0 == ret && false == file_options.fAnyOperationsAborted;
 #else
   DIR * dir = opendir(p.string().c_str());
+  if (dir == nullptr) {
+    return false;
+  }
+  RCPPUTILS_SCOPE_EXIT(closedir(dir); );
   struct dirent * directory_entry;
   while ((directory_entry = readdir(dir)) != nullptr) {
     // Make sure to not call ".." or "." entries in directory (might delete everything)
@@ -488,7 +493,6 @@ bool remove_all(const path & p)
       }
     }
   }
-  closedir(dir);
   // directory is empty now, call remove
   rcpputils::fs::remove(p);
   return !rcpputils::fs::exists(p);
--- a/src/filesystem_helper.cpp
+++ b/src/filesystem_helper.cpp
@@ -478,7 +478,7 @@ bool remove_all(const path & p)
   if (dir == nullptr) {
     return false;
   }
-  RCPPUTILS_SCOPE_EXIT(closedir(dir); );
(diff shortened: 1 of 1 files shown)


fujitatomoya (commented):
@SoftSec-Tech if you can run the static analyzer that you use to make sure leak warning is gone with this PR, that would be really appreciated.

fujitatomoya:
Pulls: ros2/rcpputils#201
Gist: https://gist.githubusercontent.com/fujitatomoya/e93ff2202151443b519a3e8087854ccf/raw/ae9709127445b675e8810a4b11f48bc3aa9c058f/ros2.repos
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14822
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22128/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16383/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1647/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22872/)

src/filesystem_helper.cpp
```diff
@@ -474,6 +475,10 @@ bool remove_all(const path & p)
   return 0 == ret && false == file_options.fAnyOperationsAborted;
 #else
   DIR * dir = opendir(p.string().c_str());
+  if (dir == nullptr) {
+    return false;
+  }
+  RCPPUTILS_SCOPE_EXIT(closedir(dir); );
```
christophebedard, line 481: I think you can just do:

```suggestion
  RCPPUTILS_SCOPE_EXIT(closedir(dir));
```

the macro already adds a `;` after the code.

src/filesystem_helper.cpp
```diff
@@ -474,6 +475,10 @@ bool remove_all(const path & p)
   return 0 == ret && false == file_options.fAnyOperationsAborted;
 #else
   DIR * dir = opendir(p.string().c_str());
+  if (dir == nullptr) {
```
christophebedard, line 478: what happened before if `dir` was `nullptr`?

fujitatomoya: that is a good question, above lines to check if the path is directory, and if not, it tries to remove that. because of this, i think it returns false with this check. but i though that null check should be added after `opendir` since there would be a racy condition.

christophebedard (approved):
Looks reasonable to me with green CI, and ideally a confirmation that this fixes the leak.

src/filesystem_helper.cpp
clalancette: I think this would have crashed previously if `opendir` ever returned `nullptr`.  In particular, `readdir` crashes on a NULL passed-in pointer, so I think if we got e.g. `EACCES`, this would have crashed.  This is a good fix to have.

fujitatomoya:
@Mergifyio backport humble jazzy

fujitatomoya:
@clalancette @christophebedard since this could be segmentation fault and memory leak, i will backport this to jazzy and humble. (no need for iron anymore, correct?)

